PR #21, opened 2021-04-05
Fix stim.TableauSimulator.measure not resizing if needed for target

Strilanc:
- Add test command to cibuildwheel configuration
- Fix unescaped unspecified encoding unicode characters in a docstring
- Drop pypy wheels from cibuildwheel tests


Strilanc:
@googlebot

PR #131, opened 2021-08-11
Basic readers

viathor:
The API is simple, but supports rather varied file formats (one vs many measure records per file, known vs unknown record length) in a consistent way.

Tests fall into three groups: counterparts to tests for writers, write-read-compare tests and additional tests for multi-record file formats.

Follow-ups:
* benchmarking tests for the readers,
* integration with `MeasureRecord`,
* support for PTB64 file format,
* read counterpart to `write_table_data`,
* possible refactoring of `write_table_data` to fit it into `MeasureRecordWriters` and similar for the read side (contingent on a fix for the performance regression this is causing).

### DIFF
--- a/src/io/measure_record_reader.cc
+++ b/src/io/measure_record_reader.cc
@@ -1,0 +1,299 @@
+/*
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "measure_record_reader.h"
+
+#include <algorithm>
+
+using namespace stim_internal;
+
+std::unique_ptr<MeasureRecordReader> MeasureRecordReader::make(FILE *in, SampleFormat input_format, size_t max_bits) {
+    switch (input_format) {
+        case SAMPLE_FORMAT_01:
+            return std::unique_ptr<MeasureRecordReader>(new MeasureRecordReaderFormat01(in, max_bits));
+        case SAMPLE_FORMAT_B8:
+            return std::unique_ptr<MeasureRecordReader>(new MeasureRecordReaderFormatB8(in, max_bits));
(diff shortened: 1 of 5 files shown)


src/io/measure_record_reader.cc
```diff
@@ -0,0 +1,284 @@
...
+
+bool MeasureRecordReaderFormat01::read_bit() {
+    if (payload == EOF) {
+        throw std::out_of_range("Attempt to read past end-of-file");
+    }
+    if (payload == '\n') {
+        throw std::out_of_range("Attempt to read past end-of-record");
+    }
+    if (payload != '0' and payload != '1') {
+        throw std::runtime_error("Character code " + std::to_string(payload) + " does not encode a bit");
```
Strilanc, line 82: Say `"Expected '0' or '1' because input format was specified as '01'."`

src/io/measure_record_reader.cc
```diff
@@ -0,0 +1,284 @@
...
+
+/// B8 format
+
+MeasureRecordReaderFormatB8::MeasureRecordReaderFormatB8(FILE *in) : in(in) {
+}
+
+size_t MeasureRecordReaderFormatB8::read_bytes(PointerRange<uint8_t> data) {
+    if (bits_available > 0) return MeasureRecordReader::read_bytes(data);
+
+    size_t k = data.ptr_end - data.ptr_start;
```
Strilanc, line 108: `data.size()`

src/io/measure_record_reader.h
```diff
@@ -0,0 +1,141 @@
...
+#include "../simd/pointer_range.h"
+#include "../simd/simd_bit_table.h"
+
+namespace stim_internal {
+
+/// Handles reading measurement data from the outside world.
+///
+/// Child classes implement the various input formats. Each file format encodes a certain number of records.
+/// Each record is a sequence of 0s and 1s. File formats B8 and R8 encode a single record. File formats 01,
+/// HITS and DETS encode any number of records. File formats 01, B8 and R8 encode the length of each record.
```
Strilanc, line 32: I think the reader should *always* know how long the record is going to be. The length should be required in order to create the reader. The 01 and DETS and other length-in-data readers use this information for *verifying* that the number of records they get is correct. The reader should be responsible for length validation; not the caller.

Strilanc (changes requested):
Details look nice, but I want to change one of the API design decisions.

viathor:
PTAL. I plan to add one more set of tests, but these depend on the interface of the DETS reader which I'd like feedback on (especially, the new ctor signature).

Broken tests are due to #133 which will be fixed once #134 is merged.

viathor:
As expected, merging #134 have caused all tests to pass.

viathor:
Friendly ping :-)

Strilanc (approved):

viathor:
\o/

PR #134, opened 2021-08-16
Fix DETS format writer's handling of multiple records

### LINKED ISSUE #133: DETS writer forgets to shoot
Currently, when one writes multiple records out to a DETS file only the first record will be prefixed with the "shot" keyword. Other records begin with the space character. For example:

```
shot D2 L1
 D0 D3 L0 L2
```

where one would expect

```
shot D2 L1
shot D0 D3 L0 L2
```

(Found in a "write-then-read" unit test for the corresponding reader.)

viathor:
Fixes #133.


Strilanc (approved):
:scream:

PR #141, opened 2021-09-14
Add support for bulk reading of measurement data

Strilanc:
- Add MeasureRecordWriter::write_bits
- Add MeasureRecordReader::read_records_into
- Fix MeasureRecordReaderFormat{R8,B8}::next_record not reading to end of current record, and also unconditionally reporting failure
- Combine `MeasureRecordReaderFormatR8::{run_length,generated}_{0s,1s}` into `buffered_{0s,1s}`
- Fix MeasureRecordReaderFormatR8 not explicitly checking for and tracking the terminal 1 bit
- Rename `read_bytes` to `read_bits_into_bytes`

### DIFF
--- a/src/stim/io/measure_record_reader.cc
+++ b/src/stim/io/measure_record_reader.cc
@@ -399,7 +399,7 @@ void MeasureRecordReaderFormatR8::buffer_data() {
         int t = getc(in);
         if (t == EOF) {
             throw std::invalid_argument(
-                "r8 data ended too early. "
+                "r8 data ended too early.  "
                 "The extracted data ended in a 1, but there was no corresponding 0x00 terminator byte for the expected "
                 "'fake encoded 1 just after the end of the data' before the input ended.");
         } else if (t != 0) {
--- a/src/stim/io/measure_record_reader.cc
+++ b/src/stim/io/measure_record_reader.cc
@@ -71,7 +71,7 @@ size_t MeasureRecordReader::read_records_into(simd_bit_table &out, bool major_in
     size_t rec = 0;
     max_shots = std::min(max_shots, out.num_major_bits_padded());
     while (rec < max_shots) {
-        size_t n = read_bytes({out[rec].u8, out[rec].u8 + out[rec].num_u8_padded()});
+        size_t n = read_bits_into_bytes({out[rec].u8, out[rec].u8 + out[rec].num_u8_padded()});
         if (n == 0) {
             break;
         }
@@ -118,13 +118,13 @@ std::unique_ptr<MeasureRecordReader> MeasureRecordReader::make(
     }
 }
 
-size_t MeasureRecordReader::read_bytes(PointerRange<uint8_t> data) {
+size_t MeasureRecordReader::read_bits_into_bytes(PointerRange<uint8_t> out_buffer) {
     if (is_end_of_record()) {
         return 0;
     }
     char result_type = current_result_type();
     size_t n = 0;
-    for (uint8_t &b : data) {
+    for (uint8_t &b : out_buffer) {
         b = 0;
         for (size_t k = 0; k < 8; k++) {
             b |= uint8_t(read_bit()) << k;
@@ -189,18 +189,18 @@ MeasureRecordReaderFormatB8::MeasureRecordReaderFormatB8(FILE *in, size_t bits_p
     : in(in), bits_per_record(bits_per_record) {
 }
 
-size_t MeasureRecordReaderFormatB8::read_bytes(PointerRange<uint8_t> data) {
+size_t MeasureRecordReaderFormatB8::read_bits_into_bytes(PointerRange<uint8_t> out_buffer) {
     if (position >= bits_per_record) {
         return 0;
     }
 
     if (bits_available > 0) {
-        return MeasureRecordReader::read_bytes(data);
+        return MeasureRecordReader::read_bits_into_bytes(out_buffer);
     }
 
-    size_t n_bits = std::min<size_t>(8 * data.size(), bits_per_record - position);
+    size_t n_bits = std::min<size_t>(8 * out_buffer.size(), bits_per_record - position);
     size_t n_bytes = (n_bits + 7) / 8;
-    n_bytes = fread(data.ptr_start, sizeof(uint8_t), n_bytes, in);
+    n_bytes = fread(out_buffer.ptr_start, sizeof(uint8_t), n_bytes, in);
     n_bits = std::min<size_t>(8 * n_bytes, n_bits);
     position += n_bits;
     return n_bits;
@@ -316,9 +316,9 @@ MeasureRecordReaderFormatR8::MeasureRecordReaderFormatR8(FILE *in, size_t bits_p
     : in(in), bits_per_record(bits_per_record) {
 }
 
-size_t MeasureRecordReaderFormatR8::read_bytes(PointerRange<uint8_t> data) {
+size_t MeasureRecordReaderFormatR8::read_bits_into_bytes(PointerRange<uint8_t> out_buffer) {
     size_t n = 0;
-    for (uint8_t &b : data) {
+    for (uint8_t &b : out_buffer) {
         b = 0;
         if (buffered_0s >= 8) {
             position += 8;
--- a/src/stim/io/measure_record_reader.cc
+++ b/src/stim/io/measure_record_reader.cc
@@ -399,7 +399,7 @@ void MeasureRecordReaderFormatR8::buffer_data() {
         int t = getc(in);
         if (t == EOF) {
             throw std::invalid_argument(
-                "r8 data ended too early.  "
+                "r8 data ended too early. "
                 "The extracted data ended in a 1, but there was no corresponding 0x00 terminator byte for the expected "
                 "'fake encoded 1 just after the end of the data' before the input ended.");
         } else if (t != 0) {
--- a/src/stim/io/measure_record_reader.cc
+++ b/src/stim/io/measure_record_reader.cc
@@ -17,7 +17,6 @@
 #include "stim/io/measure_record_reader.h"
 
 #include <algorithm>
-#include <limits.h>
 
 using namespace stim;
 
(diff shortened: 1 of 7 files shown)


src/stim/io/measure_record_reader.cc
```diff
@@ -201,8 +225,16 @@ bool MeasureRecordReaderFormatB8::read_bit() {
...
+        read_bit();
+    }
     position = 0;
-    return false;
+    int i = fgetc(in);
+    if (i == EOF) {
+        return false;
+    }
+    ungetc(i, in);
+    return true;
```
viathor, line 237: I thought that B8 files cannot contain more than one record, but I suppose with externally-supplied record size they can.

src/stim/io/measure_record_reader.cc
```diff
@@ -62,6 +60,32 @@ bool read_uint64(FILE *in, uint64_t &value, int &next) {
...
+        if (n == 0) {
+            break;
+        }
+        if (!is_end_of_record()) {
+            throw std::invalid_argument("Failed to read data. A shot contained more bits than expected.");
+        }
+        rec++;
+        if (!next_record()) {
+            break;
+        }
```
viathor, line 84: This conditional can probably be folded into the while condition.

src/stim/io/measure_record_reader.cc
```diff
@@ -201,8 +225,16 @@ bool MeasureRecordReaderFormatB8::read_bit() {
...
+    while (!is_end_of_record()) {
+        read_bit();
+    }
     position = 0;
-    return false;
+    int i = fgetc(in);
+    if (i == EOF) {
+        return false;
+    }
+    ungetc(i, in);
```
viathor, line 236: I consider ungetc to be a code smell since it usually indicates that a different and simpler solution is possible.

In this case, we should be able to use `payload` which is the reader's internal one-byte look-ahead buffer:

```c
bool MeasureRecordReaderFormatB8::next_record() {
    while (!is_end_of_record()) read_bit();
    position = 0;
    return payload != EOF;
}
```

I don't think we even need to call `maybe_update_payload` since this is already done in `is_end_of_record` which the while loop calls.

src/stim/io/measure_record_reader.cc
```diff
@@ -312,63 +341,71 @@ size_t MeasureRecordReaderFormatR8::read_bytes(PointerRange<uint8_t> data) {
...
+        read_bit();
+    }
     position = 0;
-    return false;
+    have_seen_terminal_1 = false;
+    int i = fgetc(in);
+    if (i == EOF) {
+        return false;
+    }
+    ungetc(i, in);
```
viathor, line 370: ditto

viathor (commented):
A few comments after a quick first pass. Will review again once tests pass.

src/stim/io/measure_record_reader.cc
Strilanc: `read_bit` raises an exception when it hits EOF in the middle of a record.

src/stim/io/measure_record_reader.cc
Strilanc: All of the formats can contain an unbounded list of records.

src/stim/io/measure_record_reader.cc
Strilanc: The current semantics of `next_record` don't allow you to call it at the start of reading, so it doesn't fold nicely into the loop.

If it was split into `start_record()` and `skip_rest_of_record()` then it would be possible.

src/stim/io/measure_record_reader.cc
Strilanc: Basically, in principle I agree but fixing this requires changing the read API to have a "try start record" method that returns a bool indicating whether the data continued or not, and this method has to be called as part of reading the first record. This would be a reasonable change but it is beyond the scope of this particular PR.

src/stim/io/measure_record_reader.cc
```diff
@@ -312,63 +341,71 @@ size_t MeasureRecordReaderFormatR8::read_bytes(PointerRange<uint8_t> data) {
...
-    run_length_0s = 0;
-    while (r == 0xFF) {
-        run_length_0s += 0xFF;
+
+    // Count zeroes until a one is found.
+    int r;
+    do {
         r = getc(in);
+        if (r == EOF) {
+            throw std::invalid_argument("r8 data hit end-of-file without seeing a 1 encoded at end of data.");
```
viathor, line 390: This error message is inaccurate. It suggests that a 1 is required, but it is OK for an R8 file to encode a bitstring consisting of 0s only, e.g the following is a valid R8 file

```
0xFF 0x2D
```

which encodes 300 zeros and no ones. Similarly,

```
0x00
```

is a valid R8 file encoding the empty bitstring (and neither encodes nor uses any 1s!).

Suggestion:

```c++
throw std::invalid_argument("r8 file cannot be empty and must end with byte other than 0xFF");
```

src/stim/io/measure_record_reader.cc
```diff
@@ -312,63 +341,71 @@ size_t MeasureRecordReaderFormatR8::read_bytes(PointerRange<uint8_t> data) {
...
+    if (total_data == bits_per_record) {
+        if (getc(in) != 0) {
+            throw std::invalid_argument("r8 data ending in an encoded 1 didn't include a 0x00 byte encoding the extra 1 at end of data.");
+        }
+        have_seen_terminal_1 = true;
+    } else if (total_data == bits_per_record + 1) {
+        have_seen_terminal_1 = true;
+        buffered_1s = 0;
+    } else if (total_data > bits_per_record + 1) {
+        throw std::invalid_argument("r8 data encoded a jump past the expected end of encoded data.");
```
viathor, line 407: Out of curiosity: Was the change in logic here necessary to implement the missing support for multiple records in R8? I agree that the fact that `generated_1s` was an integer might seem a little odd (since the variable can only ever be =0 or =1, unlike `generated_0s`), but that choice did simplify the logic! The new logic is more complicated. In particular, it requires the `have_seen_terminal_1` switch which leads to more branches that could hurt performance and make it harder to have good test coverage.

viathor (approved):
LGTM with some minor comments and questions.

src/stim/io/measure_record_reader.test.cc
```diff
@@ -703,3 +708,50 @@ TEST(MeasureRecordReader, FormatDets_InvalidInput) {
...
     ASSERT_THROW({ MeasureRecordReader::make(tmp, SAMPLE_FORMAT_DETS, 3); }, std::runtime_error);
 }
+
+TEST(MeasureRecordReader, read_records_into_RoundTrip) {
+    size_t n_shots = 100;
+    size_t n_results = 512 - 8;
+
+    simd_bit_table shot_maj_data = simd_bit_table::random(n_shots, n_results, SHARED_TEST_RNG());
+    simd_bit_table shot_min_data = shot_maj_data.transposed();
+    for (const auto &kv : format_name_to_enum_map) {
```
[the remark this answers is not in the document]
viathor: To clarify: by "change this" I mean I'll rewrite other existing tests to follow the loop-over-formats approach.

src/stim/io/measure_record_reader.h
```diff
@@ -90,7 +113,7 @@ struct MeasureRecordReaderFormatB8 : MeasureRecordReader {
 
     MeasureRecordReaderFormatB8(FILE *in, size_t bits_per_record);
 
-    size_t read_bytes(PointerRange<uint8_t> data) override;
+    size_t read_bits_into_bytes(PointerRange<uint8_t> out_buffer) override;
```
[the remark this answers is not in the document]
Strilanc: The reason I changed the name is because the old name made it ambiguous that the bytes argument is being written into, instead of read out of. It also was counting bits but had bytes in the name, so the result appeared to be off by a factor of 8. I would change the writer API to be more concise rather than restore the old name.

src/stim/io/measure_record_reader.test.cc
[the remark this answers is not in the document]
Strilanc: It's good to have variety in the tests.

Also, this dictionary didn't exist when you wrote the original PR, so it would have been somewhat more inconvenient. The dictionary even includes example python code that is also being tested against the internal code (via the python tests) to ensure consistency across two implementations.

src/stim/io/measure_record_reader.h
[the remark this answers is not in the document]
viathor: The first point is addressed by changing argument name (that I agree with). The second one (return value) is indeed not great. Changing writer API to match SGTM.

src/stim/io/measure_record_reader.cc
Strilanc: Made the error message clearer.

src/stim/io/measure_record_reader.cc
Strilanc: Oh, was the old logic accounting for the terminal 1? I wasn't able to see that it was doing that so I just ripped it out and did what you see here.

src/stim/io/measure_record_reader.cc
viathor: Yeah, this is definitely better. I think the error may also be thrown on an empty file, but this is very minor (and I haven't followed the full codepath for this case).

Strilanc: An empty file is invalid r8 data (assuming you expect a record to be present). I think this issue falls under having an explicit step to start a record being read.

src/stim/io/measure_record_reader.cc
viathor: Well, I might have had a bug, but I did recognize the need to do so and intended to account for it. There was some related subtleties around (re-)initializing run_length_1s related to that goal.

Anyway, I was just curious. I don't object to the new logic, especially since it handles multiple records (unlike the old logic which was written under the wrong assumption that R8 doesn't have multiple records).

src/stim/io/measure_record_reader.cc
viathor: I agree empty file is invalid for r8 (hence its explicit appearance in suggestion above). This one is OK too though.

viathor (approved):

PR #166, opened 2021-11-01
feat: add Nix/flake setup to Stim

lovesegfault:
Alright @Strilanc, a promise is a debt and here's the payment :)

This packages libstim and pystim with Nix, and makes it available for a range of
Python interpreters.

It would be nicer to do this if Stim used Poetry, but I think things are working
alright as-is.

You can build libstim and all the python packages with `nix build`, and you can
use the `overlay` attribute to easily import Stim into another flake:

```Nix
{
  inputs.stim = "github:lovesegfault/Stim/nix-init";
  outputs = { self, nixpkgs, stim }:
  let
    system = "x86_64-linux";
    pkgs = import nixpkgs { inherit system; overlays = [ stim.overlay ]; };
  in
  {
    defaultPackage = pkgs.writeShellScript "stimDemo" ''
      ${pkgs.stim}/bin/stim "${@}"
    '';
  };
}
```

I'd be happy to try my hand at moving Stim to Poetry in the future, if there's
interest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ venv/*
 .venv/*
 **/LastTest.log
 *.so
+result
+result*
+*.drv
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,64 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "locked": {
+        "lastModified": 1634851050,
+        "narHash": "sha256-N83GlSGPJJdcqhUxSCS/WwW5pksYf3VP1M13cDRTSVA=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "c91f3de5adaf1de973b797ef7485e441a65b8935",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "gitignore": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ]
+      },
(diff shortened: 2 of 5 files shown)


Strilanc (commented):
Does this use the flags from `setup.py`? It's important that `-fno-strict-aliasing` gets set, because of certain, uh, liberties taken in rather core places of the code.

Also usually build with `-std=c++11`

lovesegfault:
> Does this use the flags from `setup.py` or `CMakeLists.txt`? It's important that `-fno-strict-aliasing` gets set, because of certain, uh, liberties taken in rather core places of the code.
> 
> Also usually build with `-std=c++11`

When building the `stim` attribute this is using the CMake build, while `pystim-python3*` builds are using the setup.py.

Other than setting the correct SSE levels, no meddling with the flags set in the build systems is being done.

lovesegfault:
@Strilanc If you build the following:

```console
$ nix build -L .#stim-avx2
$ nix build -L .#stim-sse
$ nix build -L .#stim
```

You should be able to observe some build failures. I also got the gtests working.

Strilanc:
Given that this ultimately failed to produce pip-compatible wheels, I'm going to close this as "too depressing :disappointed:". But *very* open to adding nix in the future if we ever get it to work.

ChrisPattison:
I just came across this while searching for something else. You might like to know that stim is currently packaged in nix https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/stim/default.nix as of last August

PR #172, opened 2021-12-10
Explicitly include <limits> for std::numeric_limits

fvkg:


fvkg:
@Strilanc, can you do this in a separate PR?

Strilanc:
@fvkg Huh? why a separate PR?

PR #174, opened 2021-12-13
Explicitly include <limits> in detector_error_model.cc

### LINKED ISSUE #173: #include <limits>
For whatever reason, I've started to get

```stim-src/src/stim/dem/detector_error_model.cc:340:35: error: ‘numeric_limits’ is not a member of ‘std’```  

  If others have had this problem, would it be worth just importing limits to detector_error_model.cc?

maffoo:
Fixes #173


newmanmg (approved):
The hero we deserve.

PR #197, opened 2022-02-04
Use cirq.ResultDict in StimSampler when available

maffoo:
The `cirq.Result` constructor is deprecated as of https://github.com/quantumlib/Cirq/pull/4806. Instead, we should use a concrete type like `cirq.ResultDict`. This changes `StimSampler` to use `cirq.ResultDict` when available, to avoid deprecation warnings.

I'm also curious to know what is the support policy for stimcirq supporting different pyle versions. This workaround can be removed whenever support for cirq < 0.14 is dropped.

### DIFF
--- a/glue/cirq/stimcirq/_stim_sampler.py
+++ b/glue/cirq/stimcirq/_stim_sampler.py
@@ -5,6 +5,12 @@ import cirq
 from ._cirq_to_stim import cirq_circuit_to_stim_data
 
 
+try:
+    ResultImpl = cirq.ResultDict  # For cirq >= 0.14
+except AttributeError:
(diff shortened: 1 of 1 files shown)


Strilanc (approved):

Strilanc:
To be honest, stimcirq's support policy across version is "I try not to think about it too much". I make sure the latest versions of stim and stimcirq work together, and that stimcirq works with the latest stable version of cirq, and otherwise am just hoping for the best.

One of the reasons I refuse to pin the cirq version that stimcirq depends on is so that downstream users have an easier time forcing a pair that works should the worst occur.

PR #201, opened 2022-02-04
Fixing build on MacOS

noajshu:
I had trouble building Stim with tests on MacOS with AppleClang 13.0.0.13000029. These changes fixed my build.

First, `fatal error: 'gtest/gtest.h' file not found`, which is fixed by adding the explicit `GTest::` namespace prefix when calling `target_link_libraries(...)` (see also [[1]](https://cmake.org/cmake/help/latest/module/FindGTest.html)).

Second, `<array>` is not being pulled in and I get `error: implicit instantiation of undefined template 'std::array<...>` errors.
This is probably being pulled in on some other platforms by the library headers. But on MacOS the include is needed explicitly.

Third, my compiler is displeased by the move of a temporary object in the tests of PaulIString. I.e.
`error: moving a temporary object prevents copy elision [-Werror,-Wpessimizing-move]`
I just introduced an intermediate variable to work around this.

Thanks!


Strilanc:
Looks great! I'll merge it once it passes ci.

PR #205, opened 2022-02-28
add preprocessor clauses for bare support for aarch64

ecpeterson:
Stim includes some C preprocessor logic which assists in looking up processor intrinsics for specialized / accelerated operator evaluation. Currently, the logic forks on x86 Windows / x86 not-Windows, which breaks on modern Apple devices with ARM64 processors. This PR adds basic support for these processors by adding an extra (dummy) clause to the preprocessor logic, but it does not supply any specialized support for ARM64 features.

This borrows from the ARM64 clause here: https://opensource.apple.com/source/WTF/WTF-7601.1.46.42/wtf/Platform.h.auto.html .

### DIFF
--- a/src/stim/py/march.pybind.cc
+++ b/src/stim/py/march.pybind.cc
@@ -4,6 +4,14 @@
 //  Windows
 #include <intrin.h>
 #define cpuid(info, x) __cpuidex(info, x, 0)
+#elif (defined(__arm64__) && defined(__APPLE__)) || defined(__aarch64__)
+// macOS ARM64 (dummied out)
+void cpuid(int info[4], int infoType) {
+    info[0] = 0;
+    info[1] = 0;
+    info[2] = 0;
+    info[3] = 0;
+}
 #else
 //  GCC Intrinsics
 #include <cpuid.h>


ecpeterson:
I updated my email address for the commit, which auto-signed the CLI. I don't know if that sends a notification to the maintainers, so I'm also leaving this comment.

Strilanc:
Urgh, why do I get this sinking feeling of plugging ever multiplying invisible holes. I can't test for this bug myself without owning a mac, and I can't make sure it stays fixed by adding a test to CI because AFAIK [github actions has no support for m1 macs VMs](https://github.com/actions/virtual-environments/issues/2187).

Anyways, it looks good and I assume you confirmed it works on your machine. I'll merge it once the tests that-can't-even-check-this finish passing.

Strilanc (approved):

ecpeterson:
Yeah, I sympathize. I just waded through someone else's 1k lines of conditionalized headers for some cross-platform network code, and I wouldn't wish that end state on anyone.

I have indeed been using this patch locally.

PR #244, opened 2022-05-24
Stop adding dummy edges to ensure one connected component for pymatching

Strilanc:
- pymatching no longer requires the input to have just one connected component
- Use new 'fault_ids' instead of old 'qubit_id' on edges
- Add a test that invincible observables work correctly
- Loosen sinter's declared requirements

### DIFF
--- a/glue/sample/requirements.txt
+++ b/glue/sample/requirements.txt
@@ -1,7 +1,7 @@
-matplotlib == 3.5.1
-networkx == 2.6.3
-numpy == 1.22.2
-pymatching == 0.6.0
-pytest == 7.0.1
-pandas == 1.4.1
-stim == 1.8.0
+matplotlib ~= 3.5
+networkx ~= 2.6
+numpy ~= 1.22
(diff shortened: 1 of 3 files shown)


oscarhiggott:
FYI in the same PyMatching v0.4.0 release the requirement for all qubit_id's up to the max index was also removed, so the lines 
```
# Ensure invincible observables are seen by adding a boundary edge with all observables.
    g.add_node(num_detectors + 1)
    g.add_edge(num_detectors, num_detectors + 1, weight=1, qubit_id=list(range(num_observables)))
```

can also be removed

Strilanc:
Eh, it's actually corner-casier to try to only use the max anyways since it fails in the degenerate case where there's no observables.

oscarhiggott:
The number of observables (and hence shape of the array output by decode) is just the max fault id - 1. So if you want the matching.decode output to be consistent then yeah you'll want to make sure the max id is added somewhere (or reshape the decode output array if necessary). So maybe just easiest to leave it as is

PR #288, opened 2022-07-15
[draft] a little doc bug in r8 parser

balopat:
I think we only want to append the 0s to the shot when it's not the next byte is not special.

### DIFF
--- a/doc/result_formats.md
+++ b/doc/result_formats.md
@@ -431,13 +431,14 @@ events.
 *Example r8 parsing code (python)*:
 ```python
 from typing import List
+import math
 
 def parse_r8(data: bytes, bits_per_shot: int) -> List[List[bool]]:
     shots = []
     shot = []
     for byte in data:
         if byte != 255:
-            shot += [False] * byte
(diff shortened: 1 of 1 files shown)


balopat:
haha, the cla flag, I'll see what I can do.

balopat:
actually I'm not sure about this solution either, need to actually test it. Marking it as draft.

Strilanc:
I'm quite confident the original version is correct and this version is wrong. When you see `0xFF` you're supposed to append 255 zeroes. Otherwise the special byte would be a no-op and you'd have no way of encoding runs of 0s longer than 254.

These snippets are tested, by the way, so I'm expecting CI to fail.

Strilanc:
(Actually the failure will likely be that the docs need to be regenerated because you edited the output instead of the input.)

balopat:
Ah, very nice, I misunderstood on the first read, but you actually specify it clearly in the doc `it indicates to include 255 False bits but not follow them by a True bit. `. Closing!

PR #334, opened 2022-08-23
Add cmake stim_python_bindings target

danielbarter:
adding a cmake target to build the python bindings. This is mostly useful for getting a complete `compile_commands.json` out of cmake (currently we are missing all the python related headers), but also nice if you just want a python loadable shared library for testing rather than having to pip install


CMakeLists.txt
```diff
@@ -214,3 +237,13 @@ if(${GTest_FOUND})
...
 endif()
+
+find_package(Python COMPONENTS Interpreter Development)
+find_package(pybind11 CONFIG)
+if (${pybind11_FOUND} AND ${Python_FOUND})
+  pybind11_add_module(stim_python_bindings ${PYTHON_API_FILES} ${SOURCE_FILES_NO_MAIN})
+  set_target_properties(stim_python_bindings PROPERTIES OUTPUT_NAME stim)
+  add_compile_definitions(STIM_PYBIND11_MODULE_NAME=stim)
+else()
+  message("WARNING: Skipped stim_python_bindings target. `pybind11` not found")
```
Strilanc, line 248: This should provide instructions on how to fix the issue, similar to how the test target failure links to the googletest documentation.

Strilanc (changes requested):
Using cmake to make the python bindings should be added to the developer documentation.

Strilanc:
Thanks for the PR. Having a method for making the python bindings for people who use cmake instead of bazel is a good idea.

CMakeLists.txt
danielbarter: done!

Strilanc (approved):

Strilanc:
I successfully used the instructions to build it. LGTM.

Main issue is that (like the bazel one) it differs from the real package in that it does not have the AVX/SSE/nothing MUX step at the start. For similar reasons the `stim.__version__` is `VERSION_INFO`.

Strilanc:
Might have made sense to add this build step as a thing CI does, so that it doesn't break in the future. I can very easily see it breaking as `.pybind` files are added.

danielbarter:
good idea adding to CI. I can do that. 

It should be possible to get the SIMD stuff in there. Is this just a matter of propagating the correct preprocessor flags through the build process?

Strilanc:
You just overwrote my changes with your force push..

Strilanc:
The changes were to add the section to the index, and to note that you need the `out` directory in your python path to `import stim` (as well as an example command doing it and printing the version).

danielbarter:
agh, so sorry about that! Won't happen again. I can re add.

doc/developer_documentation.md
```diff
@@ -532,6 +537,26 @@ pip install -e .
...
+cmake stim_python_bindings
+# output is at out/stim.cpython-${PYTHON_VERSION}-${ARCH}.so
+```
+to import the module we need to add `out` to the python import path:
+```python
+# from the repository root
+# python
+import sys
+import os
+sys.path.append(os.getcwd() + '/out')
```
Strilanc, line 555: I would explain this as "run `PYTHON_PATH=out python -c "import stim; print(stim__version__)"`" instead of messing with python's internal sys stuff

danielbarter: ah nice. I wasn't aware that you could set is using an environment variable. Updated

PR #337, opened 2022-08-23
Bazel print files

danielbarter:
here is an approach that could be used to avoid maintaining explicit file lists in the cmake file


Strilanc:
I would rather the source of truth be the `find src | grep` commands shown in the manual build dev documentation. Also am skeptical that cmake correctly understands this, given my previous experience with trying to do anything except list-all-the-files.

PR #338, opened 2022-08-25
Move files for cmake into generated file list files

danielbarter:
Maintaining file lists directly in `CMakeList.txt` is annoying. Splitting them out into separate files which can be loaded and regenerated automatically. Changes to these files trigger a reconfigure thanks to the CMAKE_CONFIGURE_DEPENDS property


doc/developer_documentation.md
```diff
@@ -650,3 +651,16 @@ Run the following command from the repo root to auto-format all C++ code:
...
 ```
+
+# <a name="newfile"></a>Adding new C++ files
+
+For the cmake build system, we maintain C++ file lists in the folder `file_lists`.
+When you add a new C++ file, they can be updated as follows:
+
+```bash
+# from repo root
+find src | grep \.cc | grep -Ev "test|pybind|perf|^src/main\.cc" > file_lists/source_files_no_main
```
Strilanc, line 662: Add a script to the `dev/` directory that does this (in bash assuming linux is fine).

Add a step to the `ci.yml` generated-files-up-to-date check that verifies the files are up to date. (Search for "diff" in the ci.yml file).

Strilanc (changes requested):

doc/developer_documentation.md
danielbarter: done!

Strilanc (approved):

Strilanc:
Looks like the regen script might need to sort the files to ensure consistency.

dev/regen_file_lists.sh
```diff
@@ -11,6 +11,10 @@ else
     FOLDER=$1
 fi
 
+# force sorting by byte value
+export LC_ALL=C
+
+echo "forcing locale: LC_ALL=$LC_ALL"
```
Strilanc, line 17: Is it really necessary toe xport it into the calling environment?

danielbarter: yah, good point. Don't want to break something else that depends on locales

PR #344, opened 2022-09-02
Extract helper method simd_bit_table::get_index_of_bitword

danielbarter:
Small refactor of transpose code

### DIFF
--- a/src/stim/mem/simd_bit_table.h
+++ b/src/stim/mem/simd_bit_table.h
@@ -84,7 +84,7 @@ struct simd_bit_table {
     /// major_index = (major_index_high << bitword<W>::BIT_POW) + major_index_low
     /// minor_index = (minor_index_high << bitword<W>::BIT_POW) + minor_index_low
     /// As minor_index_low ranges from 0 to W-1, (*this)[major_index][minor_index] ranges over the
-    /// bits of an aligned SIMD word. get_index_of_bitword returns the index in data of
+    /// bits of an aligned SIMD word. get_index_of_bitword returns the index in data.ptr_simd of
     /// the corresponding simd word.
     inline const size_t get_index_of_bitword(size_t major_index_high, size_t major_index_low, size_t minor_index_high) const {
         size_t major_index = (major_index_high << bitword<W>::BIT_POW) + major_index_low;
--- a/src/stim/mem/simd_bit_table.h
+++ b/src/stim/mem/simd_bit_table.h
@@ -79,7 +79,7 @@ struct simd_bit_table {
     inline const simd_bits_range_ref<W> operator[](size_t major_index) const {
         return data.word_range_ref(major_index * num_simd_words_minor, num_simd_words_minor);
     }
-    /// Return the index in data of a specified row of a W * W bit block with specified major and minor index
+    /// Return the index in data of a row of a W * W bit block with given major and minor index
     inline const size_t data_index(size_t major_index, size_t minor_index, size_t row) const {
         return major_index * W * num_simd_words_minor + minor_index + row * num_simd_words_minor;
     }
--- a/src/stim/mem/simd_bit_table.h
+++ b/src/stim/mem/simd_bit_table.h
@@ -79,11 +79,11 @@ struct simd_bit_table {
     inline const simd_bits_range_ref<W> operator[](size_t major_index) const {
         return data.word_range_ref(major_index * num_simd_words_minor, num_simd_words_minor);
     }
-    /// operator[] lets us extract a specified bit as *this[major_index][minor_index].
+    /// operator[] lets us extract a specified bit as (*this)[major_index][minor_index].
     /// We can decompose these indicies as
     /// major_index = (major_index_high << bitword<W>::BIT_POW) + major_index_low
     /// minor_index = (minor_index_high << bitword<W>::BIT_POW) + minor_index_low
-    /// As minor_index_low ranges from 0 to W-1, *this[major_index][minor_index] ranges over the
+    /// As minor_index_low ranges from 0 to W-1, (*this)[major_index][minor_index] ranges over the
     /// bits of an aligned SIMD word. get_index_of_bitword returns the index in data of
     /// the corresponding simd word.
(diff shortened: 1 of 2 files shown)


src/stim/mem/simd_bit_table.h
```diff
@@ -79,6 +79,10 @@ struct simd_bit_table {
     inline const simd_bits_range_ref<W> operator[](size_t major_index) const {
         return data.word_range_ref(major_index * num_simd_words_minor, num_simd_words_minor);
     }
+    /// Return the index in data of a row of a W * W bit block with given major and minor index
+    inline const size_t data_index(size_t major_index, size_t minor_index, size_t row) const {
```
Strilanc, line 83: I don't understand what this method is computing. Part of the reason is that I explicitly avoided ever saying the word "row" or "column" in the code, since those require an ordering convention to turn into actual relevant math, so I have no idea what a row refers to here.

Rename the "row" parameter and add and "Args:" section to this method. Also this needs a way better name. "data index"? Might as well call it "number method", lol.

src/stim/mem/simd_bit_table.inl
```diff
@@ -90,12 +90,28 @@ simd_bit_table<W> simd_bit_table<W>::inverse_assuming_lower_triangular(size_t n)
     return result;
 }
 
+/// SIMD transpose:
+/// Suppose we have a matrix with minor axis contiguous in memory
+///
+/// A00  A01  ...  A0n
+/// A10  A11  ...  A1n
+///           ...
+/// An0  An1  ...  A2n
```
Strilanc, line 99: Why is the maximum index `n` for both axes? Normally the major axis has W times fewer possible indices due to the bit packing of the minor axis.

src/stim/mem/simd_bit_table.inl
```diff
@@ -90,12 +90,28 @@ simd_bit_table<W> simd_bit_table<W>::inverse_assuming_lower_triangular(size_t n)
...
+///
+/// Aij consists of W words (W * W bits), and it is stored in memory
```
Strilanc, line 101: Ah, okay, those `A` values aren't actually contiguous within themselves and actually refer to square blocks of bits. That is *very* unclear from the diagram above, though. The diagram above appears to be emphasizing that each Axy is contiguous and then contiguous with Ax(y+1).

The way I think about these operations is that each bit in the full address (down to the bit, not just down to the word) is associated with bits from the major index and bits from the minor index (again down to the bit, not down to the word). What transposing does is to adjust this mapping, so that e.g. instead of minor axis bit 0 mapping to address bit 0 the major axis bit 0 will map to address bit 0. But this may only make sense to me because I'm familiar with quantum circuit diagrams, where qubit swap operations correspond to exactly this kind of remapping within the state vector.

The key thing to realize, from an "I'm implementing bit swaps within the address bits" perspective, is that any permutation of the bits whose positions are larger than log_2(W) can be achieved via `std::swap`. But for bit positions below log_2(W) you have to start doing work within the words, and this involves things like masking and shifting within a word or between two words.

src/stim/mem/simd_bit_table.inl
```diff
@@ -105,23 +121,22 @@ void simd_bit_table<W>::do_square_transpose() {
     assert(num_simd_words_minor == num_simd_words_major);
 
     // Current address tensor indices: [...min_low ...min_high ...maj_low ...maj_high]
-
+    // transpose each Aij block
```
Strilanc, line 124: This notation isn't defined in this method?

Strilanc (changes requested):
I like extracting this method, but it needs to be better named and explained. As it stands, it is obfuscating what's happening (by hiding the details behind an unclear name that you have to jump to to unpack) instead of clarifying it.

danielbarter:
very interesting. I think I see where we are misunderstanding each other. Let me try and explain by generalizing a bit.

In my mind, a tensor with d axes consists of an array of atomic types, and a coordinate function which tells us how to map index d-tuples to data. As a warm up, here is one way to implement a tensor data structure:

```
template <size_t d, typename A>
struct tensor {
    std::array<size_t, d> bounds;
    A *data;

    size_t size() {
        size_t result = 1;
        for ( size_t n : bounds ) {
            result *= n;
        }
        return result;
    }

    A *get_data(std::array<size_t, d> coordinates) {
        size_t data_index = 0;
        size_t stride = 1;
        for (size_t i = 0; i < d; i++) {
            data_index += coordinates[i] * stride;
            stride *= bounds[i];
        }
        return data + data_index;
    }
};
```

Here, incrementing the first index corresponds to incrementing `data` by 1, so the first axis is stored contiguously in memory. If we loop over the `coordinates` array backwards in `get_data`, that would correspond to the last axis being stored contiguously in memory. The situation for Stim is a little more complex, since our atomic type isn't directly addressable, so it made sense to me to use a `bitword<W>` as the atomic type. Subconsciously, I was thinking about `simd_bit_table` as a 3-tensor with coordinates `(i,j,r)` where `(i,j)` specifies a `W bit x W bit` block and `r` specifies a row in that block. That is what the poorly named function `data_index` does: It maps those coordinates into the corresponding offset in `simd_bits`. In my mind, operations like tensor contraction and transposition live on the coordinate side, because they refer to specific indices, i.e swap the first and 3rd index.

Strilanc:
Okay, that's also how I think about it (as tensor indices), except I view each of the bits of the address as one of the indices. For example the test methods are written in terms of permuting the bits of the address. Some of the avx possibilities correspond to somewhat unusual but permutations which can be useful building blocks.

The table is a bit table so the preferred indices for the API are (major, minor) with the caveat that the bottom bits of minor are within a word.

danielbarter:
OK, mulling on that for a day, I now have a much better understanding of how you are thinking about the indexing. Updated the PR. Hopefully it is now more consistent with the surrounding code!

src/stim/mem/simd_bit_table.inl
danielbarter: have removed this now that I understand the conventions being used better

Strilanc (approved):
Yeah I find that much clearer now.

PR #346, opened 2022-09-05
Add optional seed to TableauSimulator.__init__

### LINKED ISSUE #345: Add seed=None optional argument to stim.TableauSimulator.__init__/copy
Would feed through into this method:

```
PyTableauSimulator create_tableau_simulator() {
    return PyTableauSimulator(make_py_seeded_rng(pybind11::none()));
}
```

The `copy` method should probably also take a seed argument, and potentially allow `sim.copy(seed="copy_rng")` or something similar to also duplicate the RNG state?

viathor:
Fixes #345.

### DIFF
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -252,7 +252,6 @@ API references for stable versions are kept on the [stim github wiki](https://gi
     - [`stim.TableauSimulator.c_zyx`](#stim.TableauSimulator.c_zyx)
     - [`stim.TableauSimulator.canonical_stabilizers`](#stim.TableauSimulator.canonical_stabilizers)
     - [`stim.TableauSimulator.cnot`](#stim.TableauSimulator.cnot)
-    - [`stim.TableauSimulator.copy`](#stim.TableauSimulator.copy)
     - [`stim.TableauSimulator.current_inverse_tableau`](#stim.TableauSimulator.current_inverse_tableau)
     - [`stim.TableauSimulator.current_measurement_record`](#stim.TableauSimulator.current_measurement_record)
     - [`stim.TableauSimulator.cx`](#stim.TableauSimulator.cx)
@@ -262,6 +261,7 @@ API references for stable versions are kept on the [stim github wiki](https://gi
     - [`stim.TableauSimulator.do_circuit`](#stim.TableauSimulator.do_circuit)
     - [`stim.TableauSimulator.do_pauli_string`](#stim.TableauSimulator.do_pauli_string)
     - [`stim.TableauSimulator.do_tableau`](#stim.TableauSimulator.do_tableau)
+    - [`stim.TableauSimulator.fork`](#stim.TableauSimulator.fork)
     - [`stim.TableauSimulator.h`](#stim.TableauSimulator.h)
     - [`stim.TableauSimulator.h_xy`](#stim.TableauSimulator.h_xy)
     - [`stim.TableauSimulator.h_xz`](#stim.TableauSimulator.h_xz)
@@ -7706,78 +7706,6 @@ def cnot(
     """
 ```
 
-<a name="stim.TableauSimulator.copy"></a>
-```python
-# stim.TableauSimulator.copy
-
-# (in class stim.TableauSimulator)
-def copy(
-    self,
-    *,
-    also_copy_rng_state: bool = False,
-    seed: Optional[int] = None,
-) -> stim.TableauSimulator:
-
-    """Returns a copy of the simulator. A simulator with the same internal state.
-
-    Args:
-        also_copy_rng_state: If False, RNG is reinitialized using a random seed.
-            If True, RNG state is copied together with the simulator. The two
-            copies will then produce the same measurement outcomes for the same
-            quantum circuits. If both also_copy_rng_state and seed are used,
-            an exception is raised. Defaults to False.
-        seed: PARTIALLY determines simulation results by deterministically seeding
-            the random number generator.
-
-            Must be None or an integer in range(2**64).
-
-            Defaults to None. When None, the prng is seeded from system entropy.
-
-            When set to an integer, making the exact same series calls on the exact
-            same machine with the exact same version of Stim will produce the exact
-            same simulation results.
-
-            CAUTION: simulation results *WILL NOT* be consistent between versions of
-            Stim. This restriction is present to make it possible to have future
-            optimizations to the random sampling, and is enforced by introducing
-            intentional differences in the seeding strategy from version to version.
-
-            CAUTION: simulation results *MAY NOT* be consistent across machines that
-            differ in the width of supported SIMD instructions. For example, using
-            the same seed on a machine that supports AVX instructions and one that
-            only supports SSE instructions may produce different simulation results.
-
-            CAUTION: simulation results *MAY NOT* be consistent if you vary how many
-            shots are taken. For example, taking 10 shots and then 90 shots will
-            give different results from taking 100 shots in one call.
-
-
-    Examples:
-        >>> import stim
-
-        >>> s1 = stim.TableauSimulator()
-        >>> s1.set_inverse_tableau(stim.Tableau.random(1))
-        >>> s2 = s1.copy()
-        >>> s2 is s1
-        False
-        >>> s2.current_inverse_tableau() == s1.current_inverse_tableau()
-        True
-
-        >>> s = stim.TableauSimulator()
-        >>> def brute_force_post_select(qubit, desired_result):
-        ...     global s
-        ...     while True:
-        ...         copy = s.copy()
-        ...         if copy.measure(qubit) == desired_result:
-        ...             s = copy
-        ...             break
-        >>> s.h(0)
-        >>> brute_force_post_select(qubit=0, desired_result=True)
-        >>> s.measure(0)
-        True
-    """
-```
-
 <a name="stim.TableauSimulator.current_inverse_tableau"></a>
 ```python
 # stim.TableauSimulator.current_inverse_tableau
@@ -8044,6 +7972,81 @@ def do_tableau(
         >>> sim.do_tableau(rot3, [1, 2, 3])
         >>> [str(sim.peek_bloch(k)) for k in range(4)]
         ['+Z', '+Y', '+Z', '+X']
+    """
+```
+
+<a name="stim.TableauSimulator.fork"></a>
+```python
+# stim.TableauSimulator.fork
+
+# (in class stim.TableauSimulator)
+def fork(
+    self,
+    *,
+    fresh_entropy: bool = True,
+    seed: Optional[int] = None,
+) -> stim.TableauSimulator:
+
+    """Returns a simulator with the same internal state, except perhaps its prng.
+
+    Args:
+        fresh_entropy: If True, new simulator's prng is reinitialized with a random
+            seed. If False, the prng state is copied together with the rest of the
+            original simulator's state. Consequently, in the latter case the two
+            simulators will produce the same measurement outcomes for the same
+            quantum circuits.  If both fresh_entropy and seed are set, an exception
+            is raised.  Defaults to True.
+        seed: PARTIALLY determines simulation results by deterministically seeding
+            the random number generator.
+
+            Must be None or an integer in range(2**64).
+
+            Defaults to None. When None, the prng state is either copied from the
+            original simulator or reseeded from system entropy, depending on the
+            fresh_entropy argument.
+
+            When set to an integer, making the exact same series calls on the exact
+            same machine with the exact same version of Stim will produce the exact
+            same simulation results.
+
+            CAUTION: simulation results *WILL NOT* be consistent between versions of
+            Stim. This restriction is present to make it possible to have future
+            optimizations to the random sampling, and is enforced by introducing
+            intentional differences in the seeding strategy from version to version.
+
+            CAUTION: simulation results *MAY NOT* be consistent across machines that
+            differ in the width of supported SIMD instructions. For example, using
+            the same seed on a machine that supports AVX instructions and one that
+            only supports SSE instructions may produce different simulation results.
+
+            CAUTION: simulation results *MAY NOT* be consistent if you vary how many
+            shots are taken. For example, taking 10 shots and then 90 shots will
+            give different results from taking 100 shots in one call.
+
+
+    Examples:
+        >>> import stim
+
+        >>> s1 = stim.TableauSimulator()
+        >>> s1.set_inverse_tableau(stim.Tableau.random(1))
+        >>> s2 = s1.fork()
+        >>> s2 is s1
+        False
+        >>> s2.current_inverse_tableau() == s1.current_inverse_tableau()
+        True
+
+        >>> s = stim.TableauSimulator()
+        >>> def brute_force_post_select(qubit, desired_result):
+        ...     global s
+        ...     while True:
+        ...         s2 = s.fork()
+        ...         if s2.measure(qubit) == desired_result:
+        ...             s = s2
+        ...             break
+        >>> s.h(0)
+        >>> brute_force_post_select(qubit=0, desired_result=True)
+        >>> s.measure(0)
+        True
     """
 ```
 
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -7899,9 +7899,11 @@ def __init__(
             the same seed on a machine that supports AVX instructions and one that
             only supports SSE instructions may produce different simulation results.
 
-            CAUTION: simulation results *MAY NOT* be consistent if you vary how many
-            shots are taken. For example, taking 10 shots and then 90 shots will
-            give different results from taking 100 shots in one call.
+            CAUTION: simulation results *MAY NOT* be consistent if you vary how the
+            circuit is executed. For example, reordering whether a reset on one
+            qubit happens before or after a reset on another qubit can result in
+            different measurement results being observed starting from the same
+            seed.
 
     Returns:
         An initialized stim.TableauSimulator.
@@ -8066,10 +8068,11 @@ def copy(
             the same seed on a machine that supports AVX instructions and one that
             only supports SSE instructions may produce different simulation results.
 
-            CAUTION: simulation results *MAY NOT* be consistent if you vary how many
-            shots are taken. For example, taking 10 shots and then 90 shots will
-            give different results from taking 100 shots in one call.
-
+            CAUTION: simulation results *MAY NOT* be consistent if you vary how the
+            circuit is executed. For example, reordering whether a reset on one
+            qubit happens before or after a reset on another qubit can result in
+            different measurement results being observed starting from the same
+            seed.
 
     Examples:
         >>> import stim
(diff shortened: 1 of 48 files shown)


doc/python_api_reference_vDev.md
```diff
@@ -7546,6 +7547,34 @@ class TableauSimulator:
...
+def __init__(
+    self,
+    *,
+    seed: Optional[int] = None,
+) -> None:
+
+    """Initializes a stim.TableauSimulator.
+
+    Args:
+        seed: Optional seed for the pseudo-random number generator.
```
Strilanc, line 7564: Copy the differs-across-versions warning from the other seed arguments.

doc/python_api_reference_vDev.md
```diff
@@ -7546,6 +7547,34 @@ class TableauSimulator:
...
+
+    Returns:
+        An initialized stim.TableauSimulator.
+
+    Examples:
+        >>> import stim
+        >>> s = stim.TableauSimulator(seed=0)
+        >>> s.h(0)
+        >>> s.measure(0)
+        True
```
Strilanc, line 7574: This test is not actually stable. Replace it by one that makes two simulators and checks for identical outputs.

doc/python_api_reference_vDev.md
```diff
@@ -7659,9 +7688,18 @@ def cnot(
 # (in class stim.TableauSimulator)
 def copy(
     self,
+    *,
+    copy_rng: bool = False,
 ) -> stim.TableauSimulator:
     """Returns a copy of the simulator. A simulator with the same internal state.
 
+    Args:
+        copy_rng: If False, old and new simulators share the random number
```
Strilanc, line 7697: This seems backwards to me. If I ask for the rng to be copied it should be copied, not different.

doc/python_api_reference_vDev.md
```diff
@@ -7659,9 +7688,18 @@ def cnot(
 # (in class stim.TableauSimulator)
 def copy(
     self,
+    *,
```
Strilanc, line 7691: Add an optional seed argument. Must not be specified if copy rng is specified.

Default no-arguments behavior should be new entropy for each copy. I realize this might be a breaking change (if so note it loudly), but it falls under the "fixing traps" allowance in the backwards compatibility guarantee. It's a trap for copies to all generate the same samples. That needs to be asked for.

src/stim/simulators/tableau_simulator_pybind_test.py
```diff
@@ -526,3 +526,39 @@ def test_set_state_from_stabilizers():
     assert s.current_inverse_tableau() == stim.Tableau(0)
     s.set_state_from_stabilizers([stim.PauliString("XXX"), stim.PauliString("_ZZ"), stim.PauliString("ZZ_")])
     np.testing.assert_allclose(s.state_vector(), [0.5**0.5, 0, 0, 0, 0, 0, 0, 0.5**0.5], atol=1e-6)
+
+
+def test_seed():
+    s = stim.TableauSimulator(seed=0)
+    s.h(0)
```
Strilanc, line 533: This test is not actually stable. Versions change rng seed behavior.

src/stim/simulators/tableau_simulator.pybind.cc
```diff
@@ -1475,13 +1500,25 @@ void stim_pybind::pybind_tableau_simulator_methods(pybind11::module &m, pybind11
 
     c.def(
         "copy",
-        [](const PyTableauSimulator &self) {
+        [](const PyTableauSimulator &self, bool copy_rng) {
+            if (copy_rng) {
+                return PyTableauSimulator(self, std::make_shared<std::mt19937_64>(*self.rng_reference));
+            }
             PyTableauSimulator copy = self;
```
Strilanc, line 1507: I think we need to undo the "shared rng" thing here. This is a holdover from when there was one global rng generator.

Working on the copy should not change the original in any way. Now is the time to fix these semantics since effectively thru are becoming visible for the first time.

Strilanc (changes requested):
That was fast! But gotta iterate on the semantics.

danielbarter:
Would it make more sense to implement `PyTableauSimulator` like this:
```
struct PyTableauSimulator {
    std::mt19937_64 rng;
    TableauSimulator tableau_simulator;
};
```

I can't imagine a situation where anyone would want two simulators at python level which share the same rng other than tests. Moreover, currently, a PyTableauSimulator maintains two pointers to the same rng on the heap. 

So as long as we are happy for the python tests to allocate new rng every time they allocate a new simulator, then we could have Python wrapper classes directly own the rng and still pass it to actual simulation classes as a reference.

viathor:
@danielbarter The situation where multiple components (simulators, samplers etc) share an RNG makes me uneasy, because it means that the state of those components isn't well-defined in isolation. The present issues with copying tableau simulators is a concrete practical consequence of that. There could be more issues in future, e.g. setting the state of one of those components to some fixed value could inadvertently be changing the state of other components that share the underlying RNG.

I think @Strilanc agrees since in another comment he suggests to remove RNG sharing at the C++ level. This will make `TableauSimulator` (not `PyTableauSimulator `) like `DemSampler` in terms of ownership of their RNG.

viathor:
FWIW, the `std::mt19937_64` RNG is not too heavy - about 2.5kB:

```
$ cat > r.cc
#include <random>
#include <iostream>
int main() { std::cout << sizeof(std::mt19937_64) << std::endl; return 0; }
$ g++ -o r r.cc
$ ./r
2504
$ 
```

danielbarter:
I also think that having C++ simulator/sampler classes own their rng is simpler than having wrapper classes own it. The only thing that we loose that I can see is that we woudn't be able to share rng between simulations at the C++ level, but that should really only impact the tests.

For reference, https://github.com/quantumlib/Stim/pull/352 is what it would look like to go the other way and modify DemSampler so that a wrapper class owns rng. But happy to discard

doc/python_api_reference_vDev.md
viathor: Done.

doc/python_api_reference_vDev.md
```diff
@@ -7714,20 +7714,18 @@ def cnot(
...
 
     Args:
-        fresh_entropy: If False and no seed is specified, the RNG state is copied
-            together with the rest of the simulator. The two copies will produce
-            the same measurement outcomes for the same quantum circuits. This is
-            the default behavior. If False and seed is specified, the RNG is
-            reinitialized using the new seed. If True and no seed is specified,
-            the RNG is reinitialized using a random seed. If True and seed is
-            specified, exception is raised. Defaults to False.
+        fresh_entropy: If False, the RNG state is copied together with the rest
```
Strilanc, line 7724: Rephrase so that the default option sounds passive instead of active, and the modified case sounds like an active intervention instead of a passive standing by. The user specifying this argument isn't asking for fresh entropy, that's what happens by default. The user specifying this argument wants to block fresh entropy being used, they want to keep the internal rng state. E.g. I'd call this perhaps `also_copy_prng_state`.

src/stim/simulators/tableau_simulator.pybind.cc
viathor: Agreed & done. I'll do this in other classes (seems there is only one: frame sim) as follow-up.

src/stim/simulators/tableau_simulator_pybind_test.py
viathor: Done.

viathor (commented):
Will address the new comment soon...

doc/python_api_reference_vDev.md
viathor: Done.

I phrased it in terms of how the new RNG is initialized in each case. I don't think we can make it any more passive than that since the generation of a new random seed can't be thought of as "copying" anything.

viathor:
All comments addressed. PTAL

doc/python_api_reference_vDev.md
Strilanc: > FWIW, I think fresh_entropy=False should be the default. There is nothing wrong or surprising if copies behave the same way.

**EXTREME DISAGREEMENT**. The absolute worst thing we can do to a user is have them think something is generating independent samples when it is not! This can waste months of someone's time, and then require retraction of papers! It is super duper extremely very much the case that the correct default behavior is to re-seed when copying.

Incidentally, this is why sinter uses the much more expensive "spawn" subprocess style rather than "fork". Because these kinds of mistakes terrify me.

doc/python_api_reference_vDev.md
```diff
@@ -7659,9 +7713,45 @@ def cnot(
...
+            If True, RNG state is copied together with the simulator. The two
+            copies will then produce the same measurement outcomes for the same
+            quantum circuits. If both also_copy_rng_state and seed are used,
+            an exception is raised. Defaults to False.
+        seed: PARTIALLY determines simulation results by deterministically seeding
+            the random number generator.
+
+            Must be None or an integer in range(2**64).
+
+            Defaults to None. When None, the prng is seeded from system entropy.
```
Strilanc, line 7734: If the user specifies both `alsp_copy_rng_state` and `seed`, there should be an exception. Also, this text here should say that if it's set to None then the `also_copy_rng_state` argument is controlling what happens.

src/stim/simulators/tableau_simulator.cc
```diff
@@ -715,7 +723,7 @@ void TableauSimulator::Z(const OperationData &target_data) {
     }
 }
 
-simd_bits<MAX_BITWORD_WIDTH> TableauSimulator::sample_circuit(const Circuit &circuit, std::mt19937_64 &rng, int8_t sign_bias) {
+simd_bits<MAX_BITWORD_WIDTH> TableauSimulator::sample_circuit(const Circuit &circuit, const std::mt19937_64 &rng, int8_t sign_bias) {
```
Strilanc, line 726: Is this supposed to have been changed to const? The semantics of the method are changed pretty drastically by this change. This may have broken some of the fuzz tests so that they only check one case.

This method is clearly a sortof leftover from the old way the code was using RNGs by passing them around. It probably needs to be refactored much more extremely in order to make it work with the new way where the simulators own their RNGs. Either the RNG state needs to be written back into this value as the method returns, or the parameter needs to be refactored to be like a seed or something.

src/stim/simulators/tableau_simulator.cc
```diff
@@ -734,7 +742,7 @@ void TableauSimulator::ensure_large_enough_for_qubits(size_t num_qubits) {
     inv_state.expand(num_qubits);
 }
 
-void TableauSimulator::sample_stream(FILE *in, FILE *out, SampleFormat format, bool interactive, std::mt19937_64 &rng) {
+void TableauSimulator::sample_stream(FILE *in, FILE *out, SampleFormat format, bool interactive, const std::mt19937_64 &rng) {
```
Strilanc, line 745: Is this supposed to have been changed to const? The semantics of the method are changed pretty drastically by this change. This may have broken some of the fuzz tests so that they only check one case.

This method is clearly a sortof leftover from the old way the code was using RNGs by passing them around. It probably needs to be refactored much more extremely in order to make it work with the new way where the simulators own their RNGs. Either the RNG state needs to be written back into this value as the method returns, or the parameter needs to be refactored to be like a seed or something.

Strilanc (changes requested):

doc/python_api_reference_vDev.md
viathor: Yeah, this makes sense. Thanks! However, what your argument says is that copying is a potentially very harmful operation for a sim, so should not be easily available in the API. Nevertheless, the actually useful operation - creating a new sim with "deterministic" portion of the state copied over from another sim and the RNG state reseeded with new entropy - is not a copy and naming it so is misleading.

Naming suggestions (not implemented): copy -> new_from, dup, bud, replicate... A newcomer to stim is less likely to form wrong preconceptions about what these do compared to "copy". These function names also allow us to rename the boolean arg from "also_copy_prng_state" to "fresh_entropy" or similar. The former name isn't great: it makes the reader worry that copying other parts of the sim's state may be contingent on a boolean argument, too, e.g. should this call to copy also set "also_copy_measurement_record"?

Better yet, we could merge `copy` into  `__init__` by changing the existing

```python
def __init__(self, *, seed: Optional[int] = None) -> None:
```

to

```python
def __init__(
    self,
    *,
    source: Optional[stim.TableauSimulator] = None,
    fresh_entropy: bool = True,
    seed: Optional[int] = None
) -> None:
```

WDYT?

Strilanc: We could call it `fork` instead of `copy`. I don't like merging into the init method as it overloads that method with a bunch of very task-specific arguments.

src/stim/simulators/tableau_simulator.cc
viathor: Good catch! Writing back RNG state would work for now, but seems a bit ugly and unnecessarily costly (mt19937_64 weighs about 2.5kB, see above). Also, the two functions have very few callsites, so I'd like to just fix them in this PR.

It seems to me the most elegant way to fix this is to make the two functions non-static. This simplifies function signatures and ensures that RNG initialization concerns are addressed in one place (the c'tor of TableauSimulator). The downside is that the caller has to explicitly construct the sim. However, that construction is very simple since the c'tor has only one non-optional arg (the RNG).

(Note that `reference_sample_circuit` remains static since the RNG is irrelevant.)

src/stim/simulators/tableau_simulator.cc
viathor: Will do, see other comment.

doc/python_api_reference_vDev.md
viathor: Done.

doc/python_api_reference_vDev.md
viathor: Yeah, `fork` is a very good name. I went ahead and renamed `also_copy_prng_state` to `fresh_entropy`, too.

viathor: Should we retain a variant of `copy` for backwards-compatibility?

viathor:
PTAL

doc/python_api_reference_vDev.md
```diff
@@ -249,11 +249,11 @@ API references for stable versions are kept on the [stim github wiki](https://gi
     - [`stim.TableauIterator.__iter__`](#stim.TableauIterator.__iter__)
     - [`stim.TableauIterator.__next__`](#stim.TableauIterator.__next__)
 - [`stim.TableauSimulator`](#stim.TableauSimulator)
+    - [`stim.TableauSimulator.__init__`](#stim.TableauSimulator.__init__)
     - [`stim.TableauSimulator.c_xyz`](#stim.TableauSimulator.c_xyz)
     - [`stim.TableauSimulator.c_zyx`](#stim.TableauSimulator.c_zyx)
     - [`stim.TableauSimulator.canonical_stabilizers`](#stim.TableauSimulator.canonical_stabilizers)
     - [`stim.TableauSimulator.cnot`](#stim.TableauSimulator.cnot)
-    - [`stim.TableauSimulator.copy`](#stim.TableauSimulator.copy)
```
Strilanc, line 256: Oh. I completely forgot this method already existed. You *can't* rename it because it will break backwards compatibility. My mistake, urgh.

Strilanc (changes requested):

doc/python_api_reference_vDev.md
viathor: It's ok. I made a leading change (as in "a leading question") :P

Done.

doc/python_api_reference_vDev.md
```diff
@@ -7864,6 +7865,58 @@ class TableauSimulator:
...
+            optimizations to the random sampling, and is enforced by introducing
+            intentional differences in the seeding strategy from version to version.
+
+            CAUTION: simulation results *MAY NOT* be consistent across machines that
+            differ in the width of supported SIMD instructions. For example, using
+            the same seed on a machine that supports AVX instructions and one that
+            only supports SSE instructions may produce different simulation results.
+
+            CAUTION: simulation results *MAY NOT* be consistent if you vary how many
+            shots are taken. For example, taking 10 shots and then 90 shots will
```
Strilanc, line 7903: Nit: this example could be specific to the tableau simulator, which has no shots. For example could say

`[...] *MAY NOT* be consistent if you vary how the circuit is executed. For example, reordering whether a reset on one qubit happens before or after a reset on another qubit can result in different measurement results being observed starting from the same seed.`

doc/stim.pyi
```diff
@@ -6008,8 +6053,46 @@ class TableauSimulator:
...
+                Stim. This restriction is present to make it possible to have future
+                optimizations to the random sampling, and is enforced by introducing
+                intentional differences in the seeding strategy from version to version.
+
+                CAUTION: simulation results *MAY NOT* be consistent across machines that
+                differ in the width of supported SIMD instructions. For example, using
+                the same seed on a machine that supports AVX instructions and one that
+                only supports SSE instructions may produce different simulation results.
+
+                CAUTION: simulation results *MAY NOT* be consistent if you vary how many
```
Strilanc, line 6092: Same nit about relevance of example

doc/stim.pyi
```diff
@@ -6026,9 +6109,9 @@ class TableauSimulator:
             >>> def brute_force_post_select(qubit, desired_result):
             ...     global s
             ...     while True:
```
Strilanc, line 6111: Add an example that uses the `copy_rng=True` argument. For example, just measure in the X basis in the main and the copy and assert they are the same.

src/stim/simulators/tableau_simulator.cc
```diff
@@ -715,11 +723,10 @@ void TableauSimulator::Z(const OperationData &target_data) {
     }
 }
 
-simd_bits<MAX_BITWORD_WIDTH> TableauSimulator::sample_circuit(const Circuit &circuit, std::mt19937_64 &rng, int8_t sign_bias) {
-    TableauSimulator sim(rng, circuit.count_qubits(), sign_bias);
-    sim.expand_do_circuit(circuit);
+simd_bits<MAX_BITWORD_WIDTH> TableauSimulator::sample_circuit(const Circuit &circuit) {
```
Strilanc, line 726: This method is no longer doing what it used to do. It used to give a sample from the given circuit. Now it runs the given circuit and then produces a sample. What if the simulator has already applied operations? Then it's not producing samples from the circuit. It will even include measurements run before the circuit was given.

Change this back to its old signature. Make it behave the same as it used to by moving the rng over to the simulator, and then moving it back into the argument at the end.

Strilanc (changes requested):
Okay we are getting real close now! I think this'll be the last round of comments.

src/stim/simulators/tableau_simulator.cc
viathor: Yes, it is a significant change in behavior. It seemed like what we might want to do if we are to move away from the world of shared RNGs. The previous API makes the user manage their RNGs (and hides the sims). The new one makes the user manage their sims (and hides the RNGs in the sims).

I suppose we're not moving away from the world of shared RNGs into the world of sims owning their RNGs, but instead into the world of sims that borrow and reuse RNGs as needed. Sorry, I misunderstood your intention earlier.

(Will address this and other comments soon.)

doc/python_api_reference_vDev.md
viathor: Done.

doc/stim.pyi
viathor: Done.

doc/stim.pyi
viathor: Done.

viathor:
PTAL

Strilanc (approved):
:confetti_ball:

PR #350, opened 2022-09-13
use logical or instead of bitwise or

danielbarter:
On clang 14.0.6, I am getting the following error while compiling:
```
/home/danielbarter/stim_dev/Stim/src/stim/simulators/vector_simulator.cc:159:13: error: use of bitwise '|' with boolean operands [-Werror,-Wbitwise-instead-of-logical]
        if (observable.xs[k] | observable.zs[k]) {
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                             ||
/home/danielbarter/stim_dev/Stim/src/stim/simulators/vector_simulator.cc:159:13: note: cast one or both operands to int to silence this warning
1 error generated.
```
easily fixed by using logical or instead of bitwise or

### DIFF
--- a/src/stim/simulators/vector_simulator.cc
+++ b/src/stim/simulators/vector_simulator.cc
@@ -155,11 +155,24 @@ float VectorSimulator::project(const PauliStringRef &observable) {
     };
 
     uint64_t mask = 0;
+
+    /// clang 14+ has flag -Wbitwise-instead-of-logical.
+    /// we want bitwise or here since observable.xs[k] and
+    /// observable.zs[k] are bit_refs pointing into simd words
+    #if defined __clang__ && __clang_major__ >= 14
+      #pragma clang diagnostic push
+      #pragma clang diagnostic ignored "-Wbitwise-instead-of-logical"
+    #endif
+
     for (size_t k = 0; k < observable.num_qubits; k++) {
         if (observable.xs[k] | observable.zs[k]) {
             mask |= 1ULL << k;
         }
     }
+
+    #if defined __clang__ && __clang_major__ >= 14
+      #pragma clang diagnostic pop
+    #endif
 
     basis_change();
     float mag2 = 0;


Strilanc (changes requested):
I think this warning should be suppressed instead of changing what's happening here. Keep in mind that these are AVX registers and so there's actually a bit of work going on just to turn them into a boolean in first place. And that behind the scenes work is now being doubled by doing the conversion twice. I would prefer to keep it as it is because it's safe for the compiler to turn this into the shortcut version, but it's not generally safe for it to go the other direction, or at least I don't think the compiler could figure that out from the context even though I know that it's safe.

I guess ultimately the performance difference is probably completely negligible...

danielbarter:
OK, so above error was produced by my personal laptop running clang 14. Corp laptop is also running clang 14 and there is no error, so there must be some hidden state somewhere tampering with things. 

Anyway, diffing the assembly for the two methods, they look pretty similar. Left is logical or and right is bitwise or:
```
0a1
> push rbp
6c7
< sub rsp, 0x30
---
> sub rsp, 0x38
8c9
< mov qword [rsp + 0x28], rax
---
> mov qword [rsp + 0x30], rax
16c17
< jne 0x499a43
---
> jne 0x499a54
19,20c20,21
< mov qword [rsp + 8], rsi
< mov qword [rsp + 0x10], rdi
---
> mov qword [rsp + 0x10], rsi
> mov qword [rsp + 0x18], rdi
23,24c24,25
< je 0x4998f2
< lea r15, [rsp + 0x18]
---
> je 0x49990e
> lea r15, [rsp + 0x20]
26,31c27
< jmp 0x4998ae
< nop dword [rax + rax]
< bts r13, rbx
< add rbx, 1
< cmp rbx, qword [r12]
< jae 0x4998f2
---
> nop word [rax + rax]
36,37c32,35
< mov rax, qword [rsp + 0x18]
< mov ecx, dword [rsp + 0x20]
---
> movzx ecx, byte [rsp + 0x28]
> mov ebp, 1
> shl ebp, cl
> mov rax, qword [rsp + 0x20]
39,40c37
< bt eax, ecx
< jb 0x4998a0
---
> and ebp, eax
45,46c42,43
< mov rax, qword [rsp + 0x18]
< mov ecx, dword [rsp + 0x20]
---
> mov rax, qword [rsp + 0x20]
> movzx ecx, byte [rsp + 0x28]
48c45,56
< bt eax, ecx
---
> mov edx, 1
> shl edx, cl
> and edx, eax
> mov eax, 1
> mov ecx, ebx
> shl rax, cl
> or edx, ebp
> mov ecx, 0
> cmovne rcx, rax
> or r13, rcx
> add rbx, 1
> cmp rbx, qword [r12]
50,51c58
< jmp 0x4998a4
< lea rdi, [rsp + 8]
---
> lea rdi, [rsp + 0x10]
54c61
< mov r8, qword [r14 + 8]
---
> mov rbp, qword [r14 + 8]
56,57c63,64
< cmp r8, rax
< je 0x49999e
---
> cmp rbp, rax
> je 0x4999ae
59,61c66,67
< jmp 0x49994e
< nop word cs:[rax + rax]
< nop dword [rax]
---
> jmp 0x49995e
> nop
69c75
< mov rcx, r8
---
> mov rcx, rbp
73c79
< jae 0x49999e
---
> jae 0x4999ae
92c98
< jne 0x499920
---
> jne 0x499930
96,97c102,103
< mov r8, qword [r14 + 8]
< jmp 0x49993b
---
> mov rbp, qword [r14 + 8]
> jmp 0x49994b
101c107
< jbe 0x499a62
---
> jbe 0x499a73
104,105c110,111
< movss dword [rsp + 4], xmm0
< jb 0x4999d0
---
> movss dword [rsp + 0xc], xmm0
> jb 0x4999e0
109,110c115,116
< jne 0x4999de
< jmp 0x499a15
---
> jne 0x4999ee
> jmp 0x499a25
114c120
< je 0x499a15
---
> je 0x499a25
128,129c134,135
< jb 0x4999e0
< lea rdi, [rsp + 8]
---
> jb 0x4999f0
> lea rdi, [rsp + 0x10]
132,135c138,141
< cmp rax, qword [rsp + 0x28]
< jne 0x499a81
< movss xmm0, dword [rsp + 4]
< add rsp, 0x30
---
> cmp rax, qword [rsp + 0x30]
> jne 0x499a92
> movss xmm0, dword [rsp + 0xc]
> add rsp, 0x38
140a147
> pop rbp
```
Have reverted and suppressed the warnings

Strilanc (changes requested):
Hahaha, I meant suppress it for the line, not the whole program! I do think it's a useful warning, I just consider this particular case to be a false positive.

danielbarter:
OK, have suppressed for just that loop. Compiling correctly on gcc 11, clang 13 and clang 14

Strilanc:
So, this is a bit silly, but I'm gonna merge https://github.com/quantumlib/Stim/pull/358 instead which actually includes the original change you made that I said to ignore instead. Sorry about the runaround. I do appreciate the work.

Issue #332, opened 2022-08-17
Fix the '15 detector' limit on decomposing also applying when not decomposing errors


Strilanc:
Error message says `An error case in a composite error exceeded that max supported number of symptoms`

This is only supposed to occur when `decompose_errors=True` is specified. It's limited because internally it converts the detectors affected by each of the error's possible Pauli products into bitmasks and then uses bitwise logic to understand how they mesh. When that isn't needed, it's sufficient to just compute the actual products without any bitmasking.

ChrisPattison:
I just found myself running into this limitation for `decompose_errors=False` the other day. If this isn't on your list of things to do, I could start to look into preparing a PR?

Strilanc:
That would be very helpful! Let me know if you have any troubles. That particular method is quite gnarly, but luckily the main goal is not to change the gnarly bits but to have an escape clause early on.

PR #351, opened 2022-09-13
Fix the '15 detector' limit on decomposing also applying when not decomposing errors

ChrisPattison:
Addresses #332

### DIFF
--- a/src/stim/simulators/error_analyzer.test.cc
+++ b/src/stim/simulators/error_analyzer.test.cc
@@ -2420,9 +2420,11 @@ TEST(ErrorAnalyzer, too_many_symptoms) {
     ASSERT_EQ("", check_catch<std::invalid_argument>("max supported number of symptoms", [&] {
                   ErrorAnalyzer::circuit_to_detector_error_model(symptoms_20, true, false, false, 0.0, false, true);
               }));
-    ASSERT_EQ("", check_catch<std::invalid_argument>("max supported number of symptoms", [&] {
-                  ErrorAnalyzer::circuit_to_detector_error_model(symptoms_20, false, false, false, 0.0, false, true);
-              }));
+    ASSERT_EQ(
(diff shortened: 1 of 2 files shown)


ChrisPattison:
I'm pretty unhappy with the way in which I modified the test to pass. The error should really be some `X_ERROR` annotation with an exactly representable number like `0.125`, but the first check throws a decomposition failure instead of complaining about the number of syndromes.

src/stim/simulators/error_analyzer.test.cc
```diff
@@ -2420,9 +2420,11 @@ TEST(ErrorAnalyzer, too_many_symptoms) {
     ASSERT_EQ("", check_catch<std::invalid_argument>("max supported number of symptoms", [&] {
                   ErrorAnalyzer::circuit_to_detector_error_model(symptoms_20, true, false, false, 0.0, false, true);
               }));
-    ASSERT_EQ("", check_catch<std::invalid_argument>("max supported number of symptoms", [&] {
-                  ErrorAnalyzer::circuit_to_detector_error_model(symptoms_20, false, false, false, 0.0, false, true);
-              }));
+    ASSERT_EQ(
```
Strilanc, line 2423: Copy paste this case. It should pass when not decomposing (as you've done here) and fail if decomposing (which just requires setting one of the bool arguments to true).

Strilanc (changes requested):

src/stim/simulators/error_analyzer.test.cc
ChrisPattison: There was already a case to check for a failure when `decompose_errors = True`, so I just had to modify the test to accept the expected behavior when `decompose_errors = False`. Is that what you had in mind?

Strilanc: Ah, okay, then that's fine as is.

Strilanc (approved):

PR #352, opened 2022-09-14
Dem sampler rng refactor

danielbarter:
As described in https://github.com/quantumlib/Stim/pull/346, it would be good to unify how we handle rng state so that we don't accidentally share state between python simulation objects. 

Here is the pattern that I am thinking: 
- C++ classes maintain a pointer to an rng object. 
- Python wrapper classes own an rng object which is referenced by the wrapped C++ class.

This PR is a proof of concept for DemSampler, which currently owns its own rng, in contrast to classes like TableauSimulator which just have a reference.

### DIFF
--- a/src/stim/simulators/dem_sampler.test.cc
+++ b/src/stim/simulators/dem_sampler.test.cc
@@ -31,20 +31,20 @@ TEST(DemSampler, basic_sizing) {
     ASSERT_FALSE(sampler.obs_buffer.data.not_zero());
     ASSERT_FALSE(sampler.det_buffer.data.not_zero());
 
-    sampler = DemSampler(
+    DemSampler sampler2 = DemSampler(
         DetectorErrorModel(R"DEM(
             logical_observable L2000
             detector D1000
          )DEM"),
         irrelevant_rng,
         200);
-    ASSERT_GE(sampler.det_buffer.num_major_bits_padded(), 1000);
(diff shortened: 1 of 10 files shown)


viathor (changes requested):
Leaving the decision to @Strilanc, but I think the C++ components such as samplers, simulators etc should own rather than share their RNGs. See [this comment](https://github.com/quantumlib/Stim/pull/346#issuecomment-1247406723) for rationale.

src/stim/simulators/dem_sampler.test.cc
```diff
@@ -31,20 +31,20 @@ TEST(DemSampler, basic_sizing) {
     ASSERT_FALSE(sampler.obs_buffer.data.not_zero());
     ASSERT_FALSE(sampler.det_buffer.data.not_zero());
 
-    sampler = DemSampler(
+    DemSampler sampler2 = DemSampler(
```
Strilanc, line 34: Personally I try to avoid the urge to make these sorts of small improvements to tests, because they have a nasty tendency to change what is being tested. For example, this test *could have* been checking the assignment operator but just didn't document this, and this change would have made it no longer test that.

Strilanc (changes requested):
I agree with Adam that we should instead re-architect so that each simulator owns its RNG instead of having a reference to it.

danielbarter:
OK! Closing this one

PR #354, opened 2022-09-15
Exposing FrameSimulator to python

danielbarter:
Exposing FrameSimulator to python

### DIFF
--- a/file_lists/python_api_files
+++ b/file_lists/python_api_files
@@ -13,6 +13,7 @@ src/stim/py/compiled_measurement_sampler.pybind.cc
 src/stim/py/march.pybind.cc
 src/stim/py/stim.pybind.cc
 src/stim/simulators/dem_sampler.pybind.cc
+src/stim/simulators/frame_simulator.pybind.cc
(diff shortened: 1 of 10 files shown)


Strilanc (changes requested):
I don't think this change is nearly sufficient. For example, I suspect any fuzz test is now copying a the shared RNG each time and so getting exactly the same samples.

Basically every single invocation of the constructor needs to be revisited.

I would actually recommend first changing this to std::unique_ptr, making that works (because it will make all existing calls fail), THEN inlining it to direct ownership.

danielbarter:
That is a good idea. I was planning on phasing out `SHARED_TEST_RNG` and just calling `externally_seeded_rng` directly.

Strilanc:
> I was planning on phasing out SHARED_TEST_RNG and just calling externally_seeded_rng directly. Should get to the same place right?

Yes, that's what you should do.

danielbarter:
I haven't forgotten about this! Hopefully will have some time to work on it next week

PR #356, opened 2022-09-20
Give the people filter_func

mmcewen-g:
When using sinter's convenient plotting functions, I found myself doing this a lot:
```
plotted_stats = [
            stat
            for stat in stats
            if (stat.json_metadata['b'] == 'z' and stat.json_metadata['type'] == tl)
        ]
        sinter.plot_error_rate(
            ax=ax,
            stats=plotted_stats,
            group_func=lambda stat: f"p={stat.json_metadata['p']}",
            x_func=lambda stat: stat.json_metadata['d'],
            # highlight_max_likelihood_factor=1,
        )
```
and I thought to myself, why not have sinter do that for me too?

This is a minimally invasive attempt to put the `filter_func` behavior from the command line api into the python functions.
It currently doesn't disturb the way the cmd `filter_func` is used, which is external to the plotting calls and reused between calls to `plot_error_rates` and `plot_discard_rates`. We could just hand `filter_func` down to both functions, at the cost of running this list comprehension _twice_ instead of once, but this'd make things slightly more DRY.
It also doesn't make a new function like `filter_by` to mirror `group_by` partially bc I'm lazy but mostly because it's a pretty readable single list comprehension, so that felt like a bit much.

Happy to take any of these decisions under advisement.

(I also curmudgeonly rename a variable I didn't like being reused)

### DIFF
--- a/glue/sample/src/sinter/_plotting.py
+++ b/glue/sample/src/sinter/_plotting.py
@@ -97,6 +97,7 @@ def plot_discard_rate(
         stats: 'Iterable[sinter.TaskStats]',
         x_func: Callable[['sinter.TaskStats'], Any],
         group_func: Callable[['sinter.TaskStats'], TCurveId] = lambda _: None,
+        filter_func: Callable[['sinter.TaskStats'], Any] = lambda _: True,
         plot_args_func: Callable[[int, TCurveId], Dict[str, Any]] = lambda _: {},
         highlight_max_likelihood_factor: Optional[float] = 1e3,
 ) -> None:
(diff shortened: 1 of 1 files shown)


Strilanc:
> It currently doesn't disturb the way the cmd filter_func is used, which is external to the plotting calls and reused between calls to plot_error_rates and plot_discard_rates. We could just hand filter_func down to both functions, at the cost of running this list comprehension twice instead of once, but this'd make things slightly more DRY.

Do it.

mmcewen-g:
Having now spent more than 5 seconds looking at doing it; a bunch of the nice plotting defaults (deciding whether to plot discards or not, picking a nice `min_y`) depend on having the filtered stats handy, so we'd probably compute the filtered stats here anyways?

Strilanc:
...fair enough

Strilanc (approved):

PR #441, opened 2022-11-23
Small formatting fix in gate.md readme

AlexBourassa:


Strilanc:
The change you made is actually to a file that is generated by code. You need to edit the source code that generated it. CI will fail saying you need to regen the docs, because their state disagrees with what the code makes.

AlexBourassa:
Ah!  Interesting... Didn't realize this.  I can try to track down what code, but if you have some pointers...

Strilanc:
Search for chunks of text near what you want to edit, like "`that, for backwards compatibility reasons, the targets are not combined using combin'". It should appear similarly in the source code.

Strilanc (approved):

PR #444, opened 2022-12-01
Improve parameter resolving and implemented _parameter_names_ for SweepPauli

AlexBourassa:
The current `_resolve_parameters_` would fail because the types of `new_value` and `self.cirq_sweep_symbol` would not in always be the same.  Also implemented `_parameter_names_` to expose these parameters (when for example we do `cirq.parameter_names`


AlexBourassa:
@Strilanc As requested, I added 2 unit test that would previously have failed.  Let me know what you think!

Strilanc (approved):

PR #445, opened 2022-12-07
[Kept synced for the future] Switch the targeted C++ standard from C++17 to C++20

Strilanc:


Strilanc:
More hassle to resolve merge conflicts than it will be to do this conversion again later.

PR #446, opened 2022-12-19
Fix error message with invalid sign value

ScottPJones:
Fixes a typo in the error message for `sign` attribute of a PauliString.
The allowable values are [1, -1, 1j, -1j]


Strilanc (approved):
😅

PR #450, opened 2022-12-27
Improve SVG grouping and IDs

mmcewen-g:
In order to make the SVG diagrams easier to interact with in other programs, inc graphical editors and other code, this adds groups around each slice's SVG elements, and associates unique IDs with them.

It also adds unique IDs to other elements including qubit dots and slice borders


src/stim/diagram/detector_slice/detector_slice_set.cc
```diff
@@ -615,31 +619,37 @@ void DetectorSliceSet::write_svg_diagram_to(std::ostream &out) const {
...
     for (size_t k = 0; k < num_ticks; k++) {
         for (auto q : used_qubits()) {
-            auto c = coords(min_tick + k, q);
+            auto t = min_tick + k;
+            auto qc = coordsys.qubit_coords[q]; // the raw qubit coordinates
+            auto sc = coords(t, q); // the svg coordinates, offset to the correct slice plot
             out << "<circle";
-            write_key_val(out, "cx", c.xyz[0]);
-            write_key_val(out, "cy", c.xyz[1]);
+            write_key_val(out, "id", std::format("qubitdot_{}_{}_{}", qc.xyz[0], qc.xyz[1], t))
```
Strilanc, line 626: Raw qubit coordinates aren't guaranteed to be 2 dimensional. Include a test case including a circuit with a max of qubit coordinate data, including a qubit with no coordinates, a qubit with 1d coordinates, and a qubit with 3d coordinates.

src/stim/diagram/detector_slice/detector_slice_set.cc
```diff
@@ -615,31 +619,37 @@ void DetectorSliceSet::write_svg_diagram_to(std::ostream &out) const {
...
                 out << "<rect";
+                write_key_val(out, "id", std::format("border_{}_{}_{}", row, col, k))
                 write_key_val(out, "x", sw * col * SLICE_WINDOW_GAP);
                 write_key_val(out, "y", sh * row * SLICE_WINDOW_GAP);
                 write_key_val(out, "width", sw);
                 write_key_val(out, "height", sh);
                 write_key_val(out, "stroke", "black");
                 write_key_val(out, "fill", "none");
                 out << "/>\n";
+                k++;
```
Strilanc, line 652: I think it would make sense to index using the absolute tick instead of this relative incrementing value of k.

Strilanc (changes requested):

src/stim/diagram/detector_slice/detector_slice_set.cc
mmcewen-g: `coords` above also assumes the raw coordinates has 2 elements, does that also need a check?
```
auto result = coordsys.qubit_coords[qubit];
...
result.xyz[0] += coordsys.size.xyz[0] * sx * SLICE_WINDOW_GAP;
result.xyz[1] += coordsys.size.xyz[1] * sy * SLICE_WINDOW_GAP;
return result;
```
More generally, how does the diagramming code even handle >2D qubit coords..? I figure this is just not respected in the current version.

Strilanc: The diagramming code uses an orthographic projection to turn the raw coordinate data into 2d points for the SVG canvas (or 3d in some cases if output to a model).

src/stim/diagram/detector_slice/detector_slice_set.cc
```diff
@@ -615,31 +617,59 @@ void DetectorSliceSet::write_svg_diagram_to(std::ostream &out) const {
 
     for (size_t k = 0; k < num_ticks; k++) {
         for (auto q : used_qubits()) {
-            auto c = coords(min_tick + k, q);
+            auto t = min_tick + k;
+            
+            std::stringstream id_ss;
+            id_ss << "qubitdot";
```
Strilanc, line 623: Maybe `qubit_indicator`?

src/stim/diagram/detector_slice/detector_slice_set.cc
```diff
@@ -615,31 +617,59 @@ void DetectorSliceSet::write_svg_diagram_to(std::ostream &out) const {
...
+    // Border around different slices.
     if (num_ticks > 1) {
+        size_t k = 0;
         for (uint64_t col = 0; col < num_cols; col++) {
             for (uint64_t row = 0; row < num_rows && row * num_cols + col < num_ticks; row++) {
                 auto sw = coordsys.size.xyz[0];
                 auto sh = coordsys.size.xyz[1];
+                   
+                std::stringstream id_ss;
+                id_ss << "border:" << k;
```
Strilanc, line 659: Maybe `tick_border`?

What's the reasoning for including both `k` and `k+min_tick`?

src/stim/diagram/detector_slice/detector_slice_set.cc
```diff
@@ -682,9 +713,23 @@ void DetectorSliceSet::write_svg_contents_to(
...
             if (color == nullptr) {
                 drawCorners = true;
-                color = "#AAAAAA";
+                color = BG_GREY;
             }
 
+            // Open the group element for this slice
+            out << "<g id=\"slice:" << target.val();
+            bool first = true;
+            for (auto c : detector_coordinates.at(target.val())){
```
Strilanc, line 722: Use `const auto &` instead of `auto` in loops, to avoid accidental copies and mutations. I think this loop is currently unnecessarily copy constructing every element iterated over.

src/stim/diagram/timeline/timeline_svg_drawer.cc
```diff
@@ -786,14 +786,32 @@ void DiagramTimelineSvgDrawer::make_diagram_write_to(
...
+    // Make sure qubit lines/points are drawn first, so they are in the background.
     if (mode != SVG_MODE_TIMELINE) {
         // Draw qubit points.
+        auto tick = tick_slice_start;
         for (uint64_t col = 0; col < obj.num_cols; col++) {
             for (uint64_t row = 0; row < obj.num_rows && row * obj.num_cols + col < tick_slice_num; row++) {
                 for (auto q : obj.detector_slice_set.used_qubits()) {
-                    auto c = obj.coord_sys.qubit_coords[q];
+                    std::stringstream id_ss;
+                    id_ss << "qubitdot";
```
Strilanc, line 797: Bit worried about this detail-heavy logic being repeated in two places. Extract a method?

Strilanc (changes requested):
I assume you tried working with this in inkscape and they work okay?

src/stim/diagram/detector_slice/detector_slice_set.cc
mmcewen-g: I started including k to make it easy to iterate through them, and later made it so all of them had the absolute tick at the end for similar reasons; happy to remove k if you like

src/stim/diagram/detector_slice/detector_slice_set.cc
mmcewen-g: i was thinking to label it as a visual element rather than abstractly what it does; there's other things that indicate (at) a qubit, but not other thing that is a qubit dot. Similar for qubitline rather than say qubit indicator

src/stim/diagram/detector_slice/detector_slice_set.cc
Strilanc: Na, you're the key user so you have the best sense of what's useful.

src/stim/diagram/detector_slice/detector_slice_set.cc
Strilanc: k

src/stim/diagram/diagram_util.cc
```diff
@@ -38,3 +38,16 @@ size_t stim_draw_internal::utf8_char_count(const std::string &s) {
     }
     return t;
 }
+
+void stim_draw_internal::add_vector_to_ss(std::ostream &ss, std::vector<double> vec) {
```
Strilanc, line 42: Give this a more descriptive name. I can't tell from the name why this would be colon first then underscore. Maybe "add_coord_summary_to_ss", for example

testdata/test_replacer.py
```diff
@@ -0,0 +1,21 @@
+import os
+import sys
+
+if len(sys.argv) == 1:
+    directory = os.getcwd()
+else:
+    directory = sys.argv[1]
+
+for filename in os.listdir(directory):
```
Strilanc, line 9: Don't submit this script. I want people to actually look at the images not just blindly replace them.

... I guess I could be convinced given that github has a decent comparison view. But even if it was kept this would go in the `dev/` subdirectory not `testdata/`.

Strilanc (changes requested):
Almost there! This is probably the last round.

testdata/test_replacer.py
mmcewen-g: lol I didn't mean to stage this, my bad. 
I did make it because inspecting staged changes is easier in my editor and the github comparison view is better than looking at both images when I'm expecting them not to change.

Strilanc (approved):

PR #505, opened 2023-02-23
Fix the cirq to stim conversion for circuits with tagged operations

AlexBourassa:
In particular, tagged loops will make following check fail

```python
if isinstance(op, cirq.CircuitOperation):
        self.process_circuit_operation_into_repeat_block(op)
        continue
```

Since the conversion function seems to be ignoring the tags, I simply untagged them at the beginning of the loop and let everything else proceed with the same current assumption of untagged operations.

### DIFF
--- a/glue/cirq/stimcirq/_cirq_to_stim_test.py
+++ b/glue/cirq/stimcirq/_cirq_to_stim_test.py
@@ -336,10 +336,13 @@ def test_on_loop():
                 cirq.measure(b, key="b"),
             ),
             repetitions=3,
-        )
+        ).with_tags('my_tag')
     )
     result = stimcirq.StimSampler().run(c)
     assert result.measurements.keys() == {'0:a', '0:b', '1:a', '1:b', '2:a', '2:b'}
+    
+    stim_circuit = stimcirq.cirq_circuit_to_stim_circuit(c)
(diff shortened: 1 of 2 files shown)


kjsatz:
General suggestion, is there a place we can add a test for this? Something simple with a CircuitOperation checking for these TICKs that fails before this change? Just for good practice.

Strilanc (changes requested):
Needs a unit test

glue/cirq/stimcirq/_cirq_to_stim_test.py
```diff
@@ -336,10 +336,13 @@ def test_on_loop():
                 cirq.measure(b, key="b"),
             ),
             repetitions=3,
-        )
```
Strilanc, line 339: Don't pile additional checks onto existing tests. This makes tests incomprehensible. Make a separate test.

Strilanc (changes requested):

glue/cirq/stimcirq/_cirq_to_stim_test.py
AlexBourassa: Done!

Strilanc (approved):

Issue #506, opened 2023-03-05
Add [sqrt] X axis modifier to sinter


Strilanc:
For when plotting qubit counts, but distance scales like sqrt

PR #507, opened 2023-03-06
Add a [sqrt] square root scale option to `sinter plot`

Strilanc:
- Starting an axis name with '[sqrt]' will now use a square root scale
- Heavily refactor the plot method to clarify how plots are being scaled

Fixes https://github.com/quantumlib/Stim/issues/506


Strilanc:
Note to self: add unit tests testing the sqrt data path, and leave this PR for a bit so I have the opportunity to catch plots looking funny in my dev environment since it's impossible to properly unit test this stuff.

PR #515, opened 2023-03-21
Separate repeat block in their own moment

AlexBourassa:
This modification ensures that `cirq.CircuitOperation` are always isolated in their own `cirq.Moment`

### DIFF
--- a/glue/cirq/stimcirq/_stim_to_cirq_test.py
+++ b/glue/cirq/stimcirq/_stim_to_cirq_test.py
@@ -589,3 +589,15 @@ def test_single_repeat_loops_always_flattened():
             repetitions=1,
         ),
     )) == stim.Circuit("H 0\nTICK")
+
+def test_circuit_operations_always_in_isolated_moment():
+    stim_circuit = stim.Circuit(
+        """
+        REPEAT 2 {
+            H 0 1
+        }
+        H 2 3
+        """
+    )
+    cirq_circuit = stimcirq.stim_circuit_to_cirq_circuit(stim_circuit)
+    assert len(cirq_circuit) == 2 # 1 moment would mean overlap of circuit operation
--- a/glue/cirq/stimcirq/_stim_to_cirq_test.py
+++ b/glue/cirq/stimcirq/_stim_to_cirq_test.py
@@ -601,3 +601,5 @@ def test_circuit_operations_always_in_isolated_moment():
     )
     cirq_circuit = stimcirq.stim_circuit_to_cirq_circuit(stim_circuit)
     assert len(cirq_circuit) == 2 # 1 moment would mean overlap of circuit operation
+    assert len(cirq_circuit[0].operations) == 1
+    assert isinstance(cirq_circuit[0].operations[0].untagged, cirq.CircuitOperation)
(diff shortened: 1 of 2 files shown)


Strilanc (changes requested):
Wouldn't this break round tripping of cirq circuits that do have circuit operations sharing a moment with other operations? I would consider stimtocirq(cirqtostim(A)) == A for asmanyaspossible A to be one of the most important properties for this method to have.

This change needs a unit test that would have failed but now passes.

AlexBourassa:
Well, there are actually a few reasons right now why `stimtocirq(cirqtostim(A)) != A` .  In particular,

1. The `QUBIT_COORDS` will not be consistent
2. The `DETECTOR` target ordering will not be consistent
3. Maybe other things (those were just the initial ones I noticed)

Consider the following stim circuit
```
QUBIT_COORDS(2,2) 1
QUBIT_COORDS(1,1) 0
H 0 1
H 0 1
TICK
M 0 1
DETECTOR rec[-2] rec[-1]
```

a round trip will convert it too

```
QUBIT_COORDS(1, 1) 0
QUBIT_COORDS(2, 2) 1
H 0 1
TICK
H 0 1
TICK
M 0 1
DETECTOR rec[-1] rec[-2]
TICK
```

(which is of course an equivalent circuit, but not strictly the same).

Similarly the PR here would introduce a convention of where `TICK` would go in the stim circuit, so a round trip would not give the exact same circuit (plus or minus a `TICK`), but an "equivalent circuit".

Let me know what you think!

AlexBourassa:
So in the case of starting in `cirq` and round-tripping back to `cirq` this PR would indeed break the case where we have an overlapping `cirq.CircuitOperation` with other `cirq.Operation` in the same moment.

I think this kind of `cirq` structure would actually end up breaking quite a few things in circuit translation to pulse level signal, but I'd have to defer on whether or not this is something we want to actively support in `cirq` and in `Pyle` in general.

@Strilanc Maybe worth engaging on 34113 about the issue we are currently facing.

AlexBourassa:
Ok, here is the unit test.  This basically just tests that the repeat block gets its own cirq.Moment isolated from other operations.  Let me know what you think!

glue/cirq/stimcirq/_stim_to_cirq_test.py
```diff
@@ -589,3 +589,17 @@ def test_single_repeat_loops_always_flattened():
...
+    stim_circuit = stim.Circuit(
+        """
+        REPEAT 2 {
+            H 0 1
+        }
+        H 2 3
+        """
+    )
+    cirq_circuit = stimcirq.stim_circuit_to_cirq_circuit(stim_circuit)
+    assert len(cirq_circuit) == 2 # 1 moment would mean overlap of circuit operation
```
Strilanc, line 603: Assert the exact cirq circuit you're expecting.

Strilanc (changes requested):
looks good just needs more specificity in the test

glue/cirq/stimcirq/_stim_to_cirq_test.py
AlexBourassa: Done!

Strilanc (approved):

Issue #509, opened 2023-03-15
How to correctly populate `simd_bit_table` when sampling via the C++ API


markturner289:
This is a question about the memory layout of `stim::simd_bit_table` and how to correctly populate the table with samples via the C++ API.

I find that when I use `stim::detector_samples` the rows in the simd table alternate between being populated with some lit up detectors and rows that are completely blank. The code below reproduces this with all versions of libstim I tested:

```c++
#include "stim/circuit/circuit.h"
#include "stim/gen/gen_surface_code.h"
#include "stim/simulators/detection_simulator.h"

int main()
{
    stim::CircuitGenParameters circuit_params(1, 3, "rotated_memory_x");
    circuit_params.after_clifford_depolarization = 0.11;
    circuit_params.before_measure_flip_probability = 0.01;

    const auto circuit = stim::generate_surface_code_circuit(circuit_params).circuit;

    auto rng = std::mt19937_64();
    const size_t num_shots = 20;

    const stim::DetectorsAndObservables dets_obs(circuit);
    stim::simd_bit_table<stim::MAX_BITWORD_WIDTH> sample =
        stim::detector_samples(circuit, dets_obs, num_shots, false, true, rng);

    const size_t num_dets = dets_obs.detectors.size();

    // Adds some extra space below and to the right of the relevant area of the bit table
    std::cout << sample.str(num_dets * 3, num_shots + 5) << "\n\n";

    return 0;
}
```

The stdout then shows the following:

```
11.......1.11............
.........................
.1...11.1.1.11...111.....
.........................
.1..1...1.11..1.11.1.....
.........................
......11..11..1..........
.........................
.1..11.11.111...1..1.....
.........................
.1....1......111..1......
.........................
.....1.11....1...11......
.........................
..........11.111.........
.........................
1.....1.111.1.....1......
.........................
.........................
.........................
.........................
.........................
.........................
.........................
```

If I read this correctly, the first eight odd rows contain detector data and the ninth odd row is the observable. Please could you help me understand how I have managed to introduce this alternating horizontal pattern into the bit table? My sense is that I use the C++ API incorrectly because when I inspect the bit table [here](https://github.com/quantumlib/Stim/blob/838735edc3d62b834520209095c69ca605527ef0/src/stim/py/compiled_detector_sampler.pybind.cc#L42) while executing the following python code the simd bit table is densely packed. So,

```python
import stim
stim_circuit = stim.Circuit.generated(code_task="surface_code:rotated_memory_x",
[...]

PR #519, opened 2023-03-28
Templatised simulator function VTables

markturner289:
Addresses initial steps implicated by #509 by disassociating simulator function pointers from gates and introducing independent function vtables for each simulator.

Key features
 - Gate name hashing should now only happen when parsing circuits.
 - Introduces a new `Gates` enum and a mechanism for moving from a gate name's hash to a gate's ID. Consequently, fewer places in the code now do a parsing step.
 - Gates no longer store function pointers to simulator functions. Instead this PR introduces compile time vtables for each simulator.
- The vtables are currently global. In the next step when simulators are templatised we probably wouldn’t want that, but I felt it would be better to deal with that while templatising the simulators.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 cmake_minimum_required(VERSION 3.13)
 project(stim)
 include_directories(src)
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY out)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY out)
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY out)
--- a/src/stim/circuit/circuit.cc
+++ b/src/stim/circuit/circuit.cc
@@ -216,12 +216,12 @@ DetectorsAndObservables::DetectorsAndObservables(const Circuit &circuit) {
     circuit.for_each_operation([&](const Operation &p) {
         if (p.gate->flags & GATE_PRODUCES_NOISY_RESULTS) {
             tick += p.count_measurement_results();
-        } else if (p.gate->id == gate_name_to_id("DETECTOR")) {
+        } else if (p.gate->id == static_cast<uint8_t>(Gates::DETECTOR)) {
             resolve_into(p, [&](uint64_t k) {
                 jagged_detector_data.append_tail(k);
             });
             detectors.push_back(jagged_detector_data.commit_tail());
-        } else if (p.gate->id == gate_name_to_id("OBSERVABLE_INCLUDE")) {
+        } else if (p.gate->id == static_cast<uint8_t>(Gates::OBSERVABLE_INCLUDE)) {
             size_t obs = (size_t)p.target_data.args[0];
             if (obs != p.target_data.args[0]) {
                 throw std::invalid_argument("Observable index must be an integer.");
@@ -584,7 +584,7 @@ void circuit_read_operations(Circuit &circuit, SOURCE read_char, READ_CONDITION 
         circuit_read_single_operation(circuit, c, read_char);
         Operation &new_op = ops.back();
 
-        if (new_op.gate->id == gate_name_to_id("REPEAT")) {
+        if (new_op.gate->id == static_cast<uint8_t>(Gates::REPEAT)) {
             if (new_op.target_data.targets.size() != 2) {
                 throw std::invalid_argument("Invalid instruction. Expected one repetition arg like `REPEAT 100 {`.");
             }
@@ -718,7 +718,7 @@ void stim::print_circuit(std::ostream &out, const Circuit &c, const std::string 
         }
 
         // Recurse on repeat blocks.
-        if (op.gate && op.gate->id == gate_name_to_id("REPEAT")) {
+        if (op.gate && op.gate->id == static_cast<uint8_t>(Gates::REPEAT)) {
             if (op.target_data.targets.size() == 3 && op.target_data.targets[0].data < c.blocks.size()) {
                 out << indentation << "REPEAT " << op_data_rep_count(op.target_data) << " {\n";
                 print_circuit(out, c.blocks[op.target_data.targets[0].data], indentation + "    ");
@@ -764,7 +764,7 @@ Circuit Circuit::operator*(uint64_t repetitions) const {
         return *this;
     }
     // If the entire circuit is a repeat block, just adjust its repeat count.
-    if (operations.size() == 1 && operations[0].gate->id == gate_name_to_id("REPEAT")) {
+    if (operations.size() == 1 && operations[0].gate->id == static_cast<uint8_t>(Gates::REPEAT)) {
         uint64_t old_reps = op_data_rep_count(operations[0].target_data);
         uint64_t new_reps = old_reps * repetitions;
         if (old_reps != new_reps / repetitions) {
@@ -828,7 +828,7 @@ Circuit &Circuit::operator+=(const Circuit &other) {
         PointerRange<stim::GateTarget> target_data = target_buf.take_copy(op.target_data.targets);
         OperationData op_data{arg_buf.take_copy(op.target_data.args), target_data};
         operations.push_back({op.gate, op_data});
-        if (op.gate->id == gate_name_to_id("REPEAT")) {
+        if (op.gate->id == static_cast<uint8_t>(Gates::REPEAT)) {
             assert(op.target_data.targets.size() == 3);
             target_data[0].data += block_offset;
         }
@@ -1007,7 +1007,7 @@ Circuit Circuit::py_get_slice(int64_t start, int64_t step, int64_t slice_length)
     Circuit result;
     for (size_t k = 0; k < (size_t)slice_length; k++) {
         const auto &op = operations[start + step * k];
-        if (op.gate->id == gate_name_to_id("REPEAT")) {
+        if (op.gate->id == static_cast<uint8_t>(Gates::REPEAT)) {
             result.target_buf.append_tail(GateTarget{(uint32_t)result.blocks.size()});
             result.target_buf.append_tail(op.target_data.targets[1]);
             result.target_buf.append_tail(op.target_data.targets[2]);
@@ -1072,7 +1072,7 @@ Circuit Circuit::without_noise() const {
             // Drop result flip probabilities.
             auto targets = result.target_buf.take_copy(op.target_data.targets);
             result.safe_append(*op.gate, targets, {});
-        } else if (op.gate->id == gate_name_to_id("REPEAT")) {
+        } else if (op.gate->id == static_cast<uint8_t>(Gates::REPEAT)) {
             auto args = result.arg_buf.take_copy(op.target_data.args);
             auto targets = result.target_buf.take_copy(op.target_data.targets);
             result.operations.push_back({op.gate, {args, targets}});
@@ -1091,10 +1091,10 @@ Circuit Circuit::without_noise() const {
 
 void flattened_helper(
     const Circuit &body, std::vector<double> &cur_coordinate_shift, std::vector<double> &coord_buffer, Circuit &out) {
-    const uint8_t shift = gate_name_to_id("SHIFT_COORDS");
-    const uint8_t rep = gate_name_to_id("REPEAT");
-    const uint8_t qubit_coords = gate_name_to_id("QUBIT_COORDS");
-    const uint8_t detector = gate_name_to_id("DETECTOR");
+    const uint8_t shift = static_cast<uint8_t>(Gates::SHIFT_COORDS);
+    const uint8_t rep = static_cast<uint8_t>(Gates::REPEAT);
+    const uint8_t qubit_coords = static_cast<uint8_t>(Gates::QUBIT_COORDS);
+    const uint8_t detector = static_cast<uint8_t>(Gates::DETECTOR);
     for (const auto &op : body.operations) {
         uint8_t id = op.gate->id;
         if (id == shift) {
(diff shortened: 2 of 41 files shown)


CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 cmake_minimum_required(VERSION 3.13)
 project(stim)
 include_directories(src)
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
```
Strilanc, line 18: This is a big change that should probably be done separately. You can see I have a C++20 variant that's just been sitting... because I'm terrified of merging it and breaking anything about the python packaging flow.

src/stim/circuit/circuit.cc
```diff
@@ -216,12 +216,12 @@ DetectorsAndObservables::DetectorsAndObservables(const Circuit &circuit) {
     circuit.for_each_operation([&](const Operation &p) {
         if (p.gate->flags & GATE_PRODUCES_NOISY_RESULTS) {
             tick += p.count_measurement_results();
-        } else if (p.gate->id == gate_name_to_id("DETECTOR")) {
+        } else if (p.gate->id == static_cast<uint8_t>(Gates::DETECTOR)) {
```
Strilanc, line 219: Why is the static cast necessary? The `id` field isn't the enum type?

src/stim/circuit/gate_data.h
```diff
@@ -44,7 +44,46 @@ struct ErrorAnalyzer;
...
+    // From gate_data_period_3.cc
+    C_XYZ, C_ZYX,
+    // From gate_data_period_4.cc
+    SQRT_X, SQRT_X_DAG, SQRT_Y, SQRT_Y_DAG, S, S_DAG,
+    // From gate_data_pp.cc
+    SQRT_XX, SQRT_XX_DAG, SQRT_YY, SQRT_YY_DAG, SQRT_ZZ, SQRT_ZZ_DAG,
+    // From gate_data_swaps.cc
+    SWAP, ISWAP, CXSWAP, SWAPCX, ISWAP_DAG,
+
+    /// GATE ALIASES
```
Strilanc, line 73: We shouldn't include the aliases in the enum.

For now we can keep them if it makes the change smoother, but prefix them with `ALIAS_TO_REMOVE_` or something.

CMakeLists.txt
markturner289, line 18: Happy to reverse this (and the use of `if consexpr` that requires it). Is the plan to update to cpp 20 as in #445?

src/stim/circuit/gate_data.h
```diff
@@ -76,8 +115,161 @@ inline uint8_t gate_name_to_id(const char *v, size_t n) {
     return result;
 }
 
+inline uint8_t gate_hash_to_id(uint8_t hash) noexcept {
+    switch (hash) {
```
Strilanc, line 119: *Definitely* don't construct the table this way! What if I changed the hash function?? Precompute the hash table entries by iterating over the gates and their aliases computing the hash.

I think in C++17 it might be possible to make the old `gate_name_to_id` a constexpr, in which case you could switch over it I think. At the very least this method needs to be saying `case gate_name_to_hash_id("H")` instead of `case 56`.

src/stim/circuit/gate_data.h
```diff
@@ -76,8 +115,161 @@ inline uint8_t gate_name_to_id(const char *v, size_t n) {
     return result;
 }
 
+inline uint8_t gate_hash_to_id(uint8_t hash) noexcept {
```
Strilanc, line 118: The type here should be `Gates` not `uint8_t`.

src/stim/circuit/gate_data.h
```diff
@@ -44,7 +44,46 @@ struct ErrorAnalyzer;
 constexpr uint8_t ARG_COUNT_SYGIL_ANY = uint8_t{0xFF};
 constexpr uint8_t ARG_COUNT_SYGIL_ZERO_OR_ONE = uint8_t{0xFE};
 
-inline uint8_t gate_name_to_id(const char *v, size_t n) {
+enum class Gates : uint8_t {
```
Strilanc, line 47: Rename this from `Gates` to `GateId` or `GateType`.

src/stim/circuit/gate_data.h
```diff
@@ -44,7 +44,46 @@ struct ErrorAnalyzer;
...
+    // From gate_data_annotations.cc
+    DETECTOR, OBSERVABLE_INCLUDE, TICK, QUBIT_COORDS, SHIFT_COORDS,
+    // Frome gate_data_blocks.cc
```
Strilanc, line 50: Avoid coupling the enum definition to the file structure. Say what they are instead of where they are (e.g. here it's hadamard-like gates).

src/stim/circuit/gate_data.h
```diff
@@ -44,7 +44,46 @@ struct ErrorAnalyzer;
...
+    REPEAT,
+    // From gate_data_collapsing.cc
```
Strilanc, line 52: Avoid coupling the enum definition to the file structure. Say what they are instead of where they are (e.g. here it's collapsing gates).

src/stim/circuit/gate_data.h
```diff
@@ -136,12 +328,33 @@ struct ExtraGateData {
...
+template <typename SIMULATOR>
+struct GateVTable {
+   private:
+    using sim_func_ptr_t = void (SIMULATOR::*)(const OperationData&);
+    std::array<sim_func_ptr_t, 256> funcs;
+    constexpr void add_simulator_function(Gates gate_id, sim_func_ptr_t simulator_function);
+    constexpr void add_simulator_function_annotations();
+    constexpr void add_simulator_function_blocks();
+    constexpr void add_simulator_function_collapsing();
+    constexpr void add_simulator_function_controlled();
```
Strilanc, line 340: Is this method actually needed as part of the definition? I can see that implementations of the template would want to do this kind of grouping, but it seems odd to *require* it.

src/stim/circuit/gate_data.h
```diff
@@ -136,12 +328,33 @@ struct ExtraGateData {
...
+    constexpr void add_simulator_function_hada();
+    constexpr void add_simulator_function_noisy();
+    constexpr void add_simulator_function_pauli();
+    constexpr void add_simulator_function_period_3();
+    constexpr void add_simulator_function_period_4();
+    constexpr void add_simulator_function_pp();
+    constexpr void add_simulator_function_swaps();
+
+   public:
+    sim_func_ptr_t operator[](int index) const {
```
Strilanc, line 350: index type should be `Gates`.

src/stim/simulators/error_analyzer.h
```diff
@@ -228,6 +228,10 @@ struct ErrorAnalyzer {
     /// with the implicit Z basis initialization at the start of a circuit.
     void post_check_initialization();
 
+    inline void invoke(uint8_t gate_id, const OperationData& data) {
```
Strilanc, line 231: Rename to `rev_do_gate` or something like that. Something more descriptive.

src/stim/simulators/frame_simulator.h
```diff
@@ -97,6 +97,10 @@ struct FrameSimulator {
     void reset_all_and_run(const Circuit &circuit);
     void reset_all();
 
+    inline void invoke(uint8_t gate_id, const OperationData& data) {
```
Strilanc, line 100: better name

src/stim/simulators/sparse_rev_frame_tracker.h
```diff
@@ -43,6 +43,10 @@ struct SparseUnsignedRevFrameTracker {
 
     PauliString current_error_sensitivity_for(DemTarget target) const;
 
+    inline void invoke(uint8_t gate_id, const OperationData& data) {
```
Strilanc, line 46: name

src/stim/circuit/gate_data.cc
```diff
@@ -81,25 +81,293 @@ Gate::Gate() : name(nullptr) {
...
-      best_candidate_inverse_id(gate_name_to_id(best_inverse_name)) {
+      id(static_cast<uint8_t>(gate_id)),
+      best_candidate_inverse_id(static_cast<uint8_t>(best_inverse_gate)) {
+}
+
+template <typename SIMULATOR>
+constexpr void GateVTable<SIMULATOR>::add_simulator_function_annotations() {
+    if constexpr (std::is_same<SIMULATOR, TableauSimulator>::value ||
+                  std::is_same<SIMULATOR, FrameSimulator>::value) {
+        add_simulator_function(Gates::DETECTOR, &SIMULATOR::I);
```
Strilanc, line 102: Let's break the bad cyclic dependency where `gate_data` knows *anything* about any simulator, and instead rely on each simulator knowing about gate data. In particular, imagine adding a new simulator; it should be possible to do it with ideally zero touching of `gate_data.cc`. In fact, imagine adding the new simulator with no source code access (e.g. by linking to `libstim`). Make *that* work, and then use that for the simulators inside stim. Have each simulator define each gate's method, instead of each gate defining each simulator's method.

(It's okay that adding a gate requires updating every simulator in response. Bonus points if that causes a compile time failure, but realistically a unit test failure for that simulator is sufficient.)

Strilanc (changes requested):
This is great. There are some architectural changes to make but it's quite close overall.

Something I noticed is that none of the tests were updated. It actually required no change to any test?

CMakeLists.txt
Strilanc: I am in general terrified of changing the standard. I am I think fine with the C++17 change, much more fine than the C++20 change because the manylinux standard(s) for python packages have some compatibility with C++17.

The C++20 PR is kind of just there to make sure I don't accidentally break forward compatibility, which has been an issue for some users in the past (e.g. I was using the `u8` prefix on strings wrong).

src/stim/simulators/frame_simulator.h
markturner289: I renamed the function as `do_gate`. Hopefully that's clearer?

src/stim/circuit/gate_data.h
markturner289: Agree, the big switch is not good.

I've proposed another solution where the parsing is compile time capable via constexpr and type_traits. And I've completely removed the need for the switch by assigning the compile time known hashes to the entries in the enum table. This works at compile time (even with no optimisations turned on): https://godbolt.org/z/4YYhrz7To

I concluded that the switch was too problematic to improve. I learnt that a function can't immediately proceed the `case` keyword which seems like a big limitation w.r.t. our aims.

src/stim/circuit/gate_data.h
markturner289: With the changes since this feedback, I was able to simply remove them from the enum.

src/stim/circuit/circuit.cc
```diff
@@ -1091,12 +1091,12 @@ Circuit Circuit::without_noise() const {
 
 void flattened_helper(
     const Circuit &body, std::vector<double> &cur_coordinate_shift, std::vector<double> &coord_buffer, Circuit &out) {
-    const uint8_t shift = gate_name_to_id("SHIFT_COORDS");
-    const uint8_t rep = gate_name_to_id("REPEAT");
-    const uint8_t qubit_coords = gate_name_to_id("QUBIT_COORDS");
-    const uint8_t detector = gate_name_to_id("DETECTOR");
+    const GateType shift = GateType::SHIFT_COORDS;
```
Strilanc, line 1094: Inline these now that they are constexpr.

src/stim/circuit/gate_data.h
Strilanc: Yes! It's actually constexpr all the way down to the integer! Love it.

Strilanc (changes requested):
Looking better and better. Still need to refactor the function pointers to be specified by each simulator rather than by the gate data.

src/stim/circuit/gate_data.cc
markturner289: Good point. I took two steps to address this:

1. First in f12b298 I ensured that the function pointers had to be passed into the constructor of the VTable.
2. Then in fbfff42 I associated the VTables with each simulator via composition. It's a bit harder to reason about exactly how much happens at the compile stage after this step. I also wasn't able to keep the vtable const in the `SparseUnsignedRevFrameTracker`.

Strilanc: That looks good. It's unfortunate that it makes the table at construction time, but that's not on the hot path and it avoids a lot of issues we'd run into when adding templates compared to trying to make the global.

Strilanc (changes requested):
You'll need to sign the contributor license agreement in order for me to be able to merge this. Otherwise I'm basically happy with it modulo some unit tests I'll add afterward.

markturner289:
> You'll need to sign the contributor license agreement in order for me to be able to merge this.

Strange, I completed the corporate CLA this morning. I'll look into it.

Do the CI failures look daunting to you? Most of them appear to relate to the build. I can look into reversing the C++ standard if that would help here.

Strilanc:
The CI failures just look like C++ standard flags not being passed in all the right places (e.g. the errors are things like "you used a C++14 extension"). I can handle that part of the submit because I'm familiar with fixing them. I know I can make C++20 work so C++17 can also be made to work.

Strilanc:
I made a commit getting this closer to mergeable which you can merge: https://github.com/quantumlib/Stim/commits/templatised-sim-func-vtables

Strilanc:
To fix the stimjs failure, search/replace `(sim.*op.gate->tableau_simulator_function)(op.target_data);` with `sim.do_gate(op.gate->id, op.target_data);` in `glue/javascript/tableau_simulator.js.cc`.

Strilanc:
Override CLA check based on Mark asserting in email that the CLA was signed.

Strilanc (approved):

Issue #331, opened 2022-08-11
Make `make install` install something


yurivict:

Strilanc:
Could you be more specific about what you are doing and what isn't happening?

yurivict:
There is no 'install' make target. This is because there are no ```install``` instructions in CMakeLists.txt.

Strilanc:
Oh, you mean like run a make command to install libstim as a library that can be linked to or to install stim as a command line tool in the system? I've just been building it and then adding the out directory to my PATH, or using the fact that the python package includes the command line tool.

yurivict:
Library, headers. and possibly cmake and pkgconfig files.

PR #538, opened 2023-04-27
Allow cmake install

fdmalone:
Addresses #331. make install now installs the stim binary and benchmark tool to bin, libstim to lib and the headers to include. The install prefix can be controlled by CMAKE_INSTALL_PREFIX.


.github/workflows/ci.yml
```diff
@@ -240,6 +240,18 @@ jobs:
     - run: echo -e '#include "stim.h"\nint main(int argc,const char **argv) {return !stim::find_bool_argument("test", argc, argv);}' > test.cc
     - run: g++ -std=c++17 test.cc out/libstim.a -I src
     - run: ./a.out test
+  build_lib_install:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - run: mkdir install_dir
```
Strilanc, line 247: Add a documentation version of this check's contents to doc/developer_documentation.md

Strilanc (changes requested):
Nice to finally have this fixed. Looks good except it needs to be documented in the dev docs.

Strilanc (approved):
I'd have made it its own section but this is close enough; the key information is there.

Thanks for this! I was definitely gonna push this one off forever, so very nice that it got done.

PR #548, opened 2023-05-01
Templatise simulators with SIMD width NTTP

### LINKED ISSUE #509: How to correctly populate `simd_bit_table` when sampling via the C++ API

markturner289:
- Change affects the `FrameSimulator` and `TableauSimulator`.
- Introduces `.inl` files for `.cc` files that have been templatised.
- Pybind components, tests and benchmarking code only ever templatise with `MAX_BITWORD_WIDTH` matching previous behaviour. This is still safe as a user of the python API will never be in a position to accidently mix and match SIMD widths.
- The user facing functions of the libstim C++ API that return samples from Stim have been templatised so that the compiler will enforce a matching between `simd_bit_table` widths and the internal library code. This prevents bugs that can occur when the SIMD width is inferred to be different in user space than in libstim (for example, when user code is compiled without `-march=native` as used in Stim).
- Separate out `force_streaming.h/cc` to avert multiple definitions error which templatising would otherwise introduce.

Builds on #519, #556, #558 and #559 to close #509.


src/stim/simulators/force_streaming.h
```diff
@@ -0,0 +1,11 @@
+#include <cstdint>
+
+namespace stim {
+
+bool should_use_streaming_because_bit_count_is_too_large_to_store(uint64_t result_count);
```
Strilanc, line 5: While we're here, add a comment saying that this is here for testing the streaming without having to make enormous circuits.

src/stim/simulators/frame_simulator.perf.cc
```diff
@@ -26,7 +26,7 @@ BENCHMARK(FrameSimulator_depolarize1_100Kqubits_1Ksamples_per1000) {
     size_t num_samples = 1000;
     double probability = 0.001;
     std::mt19937_64 rng(0);  // NOLINT(cert-msc51-cpp)
-    FrameSimulator sim(stats, FrameSimulatorMode::STORE_DETECTIONS_TO_MEMORY, num_samples, rng);
+    FrameSimulator<MAX_BITWORD_WIDTH> sim(stats, FrameSimulatorMode::STORE_DETECTIONS_TO_MEMORY, num_samples, rng);
```
Strilanc, line 29: This is fine for now. I am definitely gonna do some sort of follow-up setting the width to various fixed numbers.

src/stim/simulators/frame_simulator.test.cc
```diff
@@ -26,7 +26,7 @@ TEST(FrameSimulator, get_set_frame) {
     CircuitStats circuit_stats;
     circuit_stats.num_qubits = 6;
     circuit_stats.max_lookback = 999;
-    FrameSimulator sim(circuit_stats, FrameSimulatorMode::STORE_DETECTIONS_TO_MEMORY, 4, SHARED_TEST_RNG());
+    FrameSimulator<MAX_BITWORD_WIDTH> sim(circuit_stats, FrameSimulatorMode::STORE_DETECTIONS_TO_MEMORY, 4, SHARED_TEST_RNG());
```
Strilanc, line 29: Use the macro `TEST_EACH_WORD_SIZE_W` instead of `TEST` (see `simd_word.test.cc` for examples), then replace `MAX_BITWORD_WIDTH` with `W`.

src/stim/simulators/frame_simulator_util.test.cc
```diff
@@ -54,22 +54,22 @@ TEST(DetectionSimulator, sample_batch_detection_events_writing_results_to_disk)
     )circuit");
 
     FILE *tmp = tmpfile();
-    sample_batch_detection_events_writing_results_to_disk(
+    sample_batch_detection_events_writing_results_to_disk<MAX_BITWORD_WIDTH>(
```
Strilanc, line 57: `TEST_EACH_WORD_SIZE_W`

src/stim/simulators/measurements_to_detection_events.test.cc
```diff
@@ -375,7 +375,7 @@ TEST(measurements_to_detection_events, file_01_to_dets_no_obs) {
     rewind(in);
     FILE *out = tmpfile();
 
-    stream_measurements_to_detection_events(
+    stream_measurements_to_detection_events<MAX_BITWORD_WIDTH>(
```
Strilanc, line 378: `TEST_EACH_WORD_SIZE_W`

Ultimately there are gonna be very few uses of `MAX_BITWORD_WIDTH` anywhere in the code base, other than the python bindings.

src/stim/simulators/tableau_simulator.test.cc
```diff
@@ -39,7 +39,7 @@ struct OpDat {
 };
 
 TEST(TableauSimulator, identity) {
-    auto s = TableauSimulator(SHARED_TEST_RNG(), 1);
+    auto s = TableauSimulator<MAX_BITWORD_WIDTH>(SHARED_TEST_RNG(), 1);
```
Strilanc, line 42: `TEST_EACH_WORD_SIZE_W`

Strilanc (changes requested):
Awesome.

My main feedback is you should use `TEST_EACH_WORD_SIZE_W` to test each word size, instead of hard coding the max word size in tests.

markturner289:
> My main feedback is you should use `TEST_EACH_WORD_SIZE_W` to test each word size, instead of hard coding the max word size in tests.

I really like this suggestion.

However, I started to implement it and found it had massive implications across large sections of the code base. Large parts of the core infrastructure defer to `MAX_BITWORD_WIDTH` so tests that parameterise simulators with SIMD widths that are less than that would require a lot of components to be templatised accordingly. This then incurs a lot of knock on templatisation if components are coupled with a template - this includes `Tableau`, `PauliStringRef`, `CommutingPauliStringIterator`, `MeasureRecordBatch` but certainly extends much further than this.

As I say though, I think the suggestion is great and I'd be interested in working on this in future. Would a follow up PR sound ok to you for this work package?

Strilanc:
@markturner289 Unfortunately, the `TEST_EACH_WORD_SIZE_W` is non-negotiable.

It's okay if, under the hood, some of the intermediate classes are using the max size. But we need to check that the correct answer is still coming out. And we need to check that anyone linking to the latest version won't have a nasty surprise.

I'd advise against trying to turn over the whole code base all at once in one PR. It's okay if you do PauliString in one PR then the frame simulator in another then the Tableau class in another then the tableau simulator in another. It's okay that that will make the number of "max_word" in the code base go way up, before it comes back back down. But whatever templates we expose along the way should be tested at not-just-the-max-word-size because using them at other word sizes is a core goal of the PR.

markturner289:
> @markturner289 Unfortunately, the TEST_EACH_WORD_SIZE_W is non-negotiable.


Fully understand.

> It's okay if, under the hood, some of the intermediate classes are using the max size.


I think this will be the key challenge for me going forward. My sense is that no intermediate classes can use `MAX_BITWORD_WIDTH` if the goal of testing templatised simulators with `TEST_EACH_WORD_SIZE_W` is to be reached. The compiler doesn't allow the max width to coexist with the differing template parameters that filter through from the simulator at "test" time. 

> I'd advise against trying to turn over the whole code base all at once in one PR.


Agreed. I think the best way forward is for me to close this PR and proceed by templatising lower level components, gradually removing the hardcoded parameterisations with `MAX_BITWORD_WIDTH` under `src/stim` across a few separate PRs (obviously there are key exceptions to this such as the pybind code).

My worry is that templatisation might end up having to touch a significant majority of classes under `src/stim`. Please can you confirm whether this still sounds like a reasonable direction of travel to you?

Strilanc:
I knew going in that it would probably turn a lot of the code base into templates. So yes I'm fine with that.

I don't necessarily agree that you need to close this PR. Just retitle it "[blocked]" until it can actually land.

markturner289:
I want to give the diff another look as the rebase was quite challenging. But I think I have some progress and can mark this unblocked.

With #556, #558 and #559 it was possible to update this PR to unit test the `FrameSimulator` and `TableauSimulator` across all SIMD widths.

To my eyes, one issue remains - please see the comment I have posted alongside the relevant code.

src/stim/simulators/tableau_simulator.test.cc
```diff
@@ -457,55 +458,56 @@ TEST(TableauSimulator, to_state_vector_canonical) {
...
+//         ASSERT_TRUE(vec_sim_corroborates_measurement_process(state, {0, 1, 2, 3}));
+//     }
+//     {
+//         auto state = Tableau<W>::random(12, SHARED_TEST_RNG());
+//         ASSERT_TRUE(vec_sim_corroborates_measurement_process(state, {0, 1, 2, 3}));
+//         ASSERT_TRUE(vec_sim_corroborates_measurement_process(state, {0, 10, 11}));
+//         ASSERT_TRUE(vec_sim_corroborates_measurement_process(state, {11, 5, 7}));
+//         ASSERT_TRUE(vec_sim_corroborates_measurement_process(state, {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11}));
+//     }
+// })
```
markturner289, line 503: I intend to give this code further thought.

The issue here is `tableau_simulator.h` includes vector simulator. This is problematic for us here because `tableau_simulator.h` doesn't instantiate template member functions that `tableau_simulator.test` subsequently needs from vector_simulator.h. Once the test file gets round to including the vector simulator, all its template member function definitions are ifdef'd out.

I need to find a way to allow the `TableauSimulator` to coexist with the `VectorSimulator` in single translation unit.

Strilanc: I don't think I understand what the issue is here. What's the actual compile error that occurs?

markturner289:
```
/home/markturner289/code/Stim-fork/src/stim/simulators/tableau_simulator.test.cc: In instantiation of ‘bool vec_sim_corroborates_measurement_process(const stim::Tableau<W>&, const std::vector<unsigned int>&) [with long unsigned int W = 256]’:
/home/markturner289/code/Stim-fork/src/stim/simulators/tableau_simulator.test.cc:483:1:   required from here
/home/markturner289/code/Stim-fork/src/stim/simulators/tableau_simulator.test.cc:474:34: error: invalid operands of types ‘<unresolved overloaded function type>’ and ‘long unsigned int’ to binary ‘operator<’
  474 |         float f = vec_sim.project<W>(buf);
      |                   ~~~~~~~~~~~~~~~^~
/home/markturner289/code/Stim-fork/src/stim/simulators/tableau_simulator.test.cc: In instantiation of ‘bool vec_sim_corroborates_measurement_process(const stim::Tableau<W>&, const std::vector<unsigned int>&) [with long unsigned int W = 128]’:
/home/markturner289/code/Stim-fork/src/stim/simulators/tableau_simulator.test.cc:483:1:   required from here
/home/markturner289/code/Stim-fork/src/stim/simulators/tableau_simulator.test.cc:474:34: error: invalid operands of types ‘<unresolved overloaded function type>’ and ‘long unsigned int’ to binary ‘operator<’
/home/markturner289/code/Stim-fork/src/stim/simulators/tableau_simulator.test.cc: In instantiation of ‘bool vec_sim_corroborates_measurement_process(const stim::Tableau<W>&, const std::vector<unsigned int>&) [with long unsigned int W = 64]’:
/home/markturner289/code/Stim-fork/src/stim/simulators/tableau_simulator.test.cc:483:1:   required from here
/home/markturner289/code/Stim-fork/src/stim/simulators/tableau_simulator.test.cc:474:34: error: invalid operands of types ‘<unresolved overloaded function type>’ and ‘long unsigned int’ to binary ‘operator<’
make[2]: *** [CMakeFiles/stim_test_o3.dir/build.make:1987: CMakeFiles/stim_test_o3.dir/src/stim/simulators/tableau_simulator.test.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:86: CMakeFiles/stim_test_o3.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

markturner289: Finally worked it out.

Where previously, in `vec_sim_corroborates_measurement_process`, I had 

```C++
        float f = vec_sim.project<W>(buf);
```

I instead needed

```C++
        float f = vec_sim.template project<W>(buf);
```

Without `.template` the type of `vec_sim` was assumed to be a non-template.

I wasn't familiar with this syntax / theory, so found this blog post on dependant types fascinating: http://www.aerialmantis.co.uk/blog/2017/03/17/template-keywords/.

src/stim/py/stim.pybind.cc
```diff
@@ -111,7 +111,7 @@ pybind11::object raw_format_data_solo(const FileFormatData &data) {
     result["parse_example"] = data.help_python_parse;
     result["save_example"] = data.help_python_save;
     result["help"] = data.help;
-    return result;
+    return std::move(result);
```
Strilanc, line 114: Does this move do anything that wouldn't happen automatically?

Strilanc (approved):
Awesome. That's a lot of work, but it should finally make that silent failure you run into impossible. And it puts the codebase in a much better place for other changes, like e.g. conditionally using smaller sizes from python when working on small objects.

Strilanc:
Because this is a pretty complex change, and there's some comments saying you wanted to check some things, I'm going to wait for you to confirm the PR is ready to merge.

src/stim/py/stim.pybind.cc
markturner289: Clang12 brought this to my attention - copy elision won't happen since the return type is of the function is different to the type of `result`.

markturner289:
Great, thanks for your patience with this PR!

I can confirm that, from my side, the PR is ready to merge.

I think this PR is complete and contained, but have a few follow up PRs in mind:
 - this PR focusses on templatising the `TableauSimulator` and `FrameSimulator`. But there are also some functions in the `ErrorAnalyzer` and the `SparseUnsignedRevFrameTracker` that could be templatised and the `DemSampler` could become a template class. After this, `MAX_BITWORD_WIDTH` will be completely isolated from the core C++ components.
 - As mentioned in #559, a last few tests in `conversions.test.cc` could be be moved over to use `TEST_EACH_WORD_SIZE_W` once this PR is in.

Strilanc:
Great! Merging it.

ErrorAnalyzer and SparseUnsignedRevFrameTracker are likely not great candidates for becoming fully templatized as they use sparse representations instead of dense simd-heavy representations. DemSampler makes more sense to me.

One possible follow-up is to modify the pybinding code for tableaus to use a smaller size when the tableau is small, e.g. add an 8 bit simd word and if the tableau is less than 64x64 then use the 8 bit one. That would involve finding some of the places where it's assumed all words are at least 64 bits wide, though.

Issue #532, opened 2023-04-24
Add `stim.reference_sample` method


Strilanc:
Basically this just calls `TableauSimulator::reference_sample_circuit` and returns the result. It's a method that returns a consistent (within one version of stim) and valid sample from the circuit, ignoring all noise processes.

fdmalone:
Am I right that this and #531 are connected in that you would like to be able to pass a reference_sample into Compiled(Measurement)Sampler.__init__?

Strilanc:
@fdmalone Yes, that's right. Giving the user control over the reference sample enables emulation of some advanced use cases, where the reference sample is not consistent from shot to shot in circuits with controlled cliffords.

PR #549, opened 2023-05-01
Allow reference_sample to be passed as an argument to sampler.

fdmalone:
Addresses #531 and #532. compile_sampler now accepts reference_sample keyword argument which defaults to None. Also exposes the TableauSimulator::reference_sample method to pybind which is useful for checking that passing in this reference sample reproduces to current behaviour when skip_reference_sample is False.

### DIFF
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -41,6 +41,7 @@ API references for stable versions are kept on the [stim github wiki](https://gi
     - [`stim.Circuit.num_qubits`](#stim.Circuit.num_qubits)
     - [`stim.Circuit.num_sweep_bits`](#stim.Circuit.num_sweep_bits)
     - [`stim.Circuit.num_ticks`](#stim.Circuit.num_ticks)
+    - [`stim.Circuit.reference_sample`](#stim.Circuit.reference_sample)
     - [`stim.Circuit.search_for_undetectable_logical_errors`](#stim.Circuit.search_for_undetectable_logical_errors)
     - [`stim.Circuit.shortest_graphlike_error`](#stim.Circuit.shortest_graphlike_error)
     - [`stim.Circuit.to_file`](#stim.Circuit.to_file)
@@ -1040,6 +1041,7 @@ def compile_sampler(
     *,
     skip_reference_sample: bool = False,
     seed: object = None,
+    reference_sample: object = None,
 ) -> stim.CompiledMeasurementSampler:
     """Returns an object that can quickly batch sample measurements from the circuit.
 
@@ -1083,6 +1085,18 @@ def compile_sampler(
             CAUTION: simulation results *MAY NOT* be consistent if you vary how many
             shots are taken. For example, taking 10 shots and then 90 shots will
             give different results from taking 100 shots in one call.
+        reference_sample: The data to xor into the measurement flips produced by the
+            frame simulator, in order to produce proper measurement results.
+            This can either be specified as an `np.bool_` array or a bit packed
+            `np.uint8` array (little endian). Under normal conditions, the reference
+            sample should be a valid noiseless sample of the circuit, such as the
+            one returned by `circuit.reference_sample()`. If this argument is not
+            provided, the reference sample will be set to
+            `circuit.reference_sample()`, unless `skip_reference_sample=True`
+            is used, in which case it will be set to all-zeros.
+
+    Raises:
+        ValueError: skip_reference_sample is True and reference_sample is not None.
 
     Examples:
         >>> import stim
@@ -1979,6 +1993,30 @@ def num_ticks(
         ...    }
         ... ''').num_ticks
         101
+    """
+```
+
+<a name="stim.Circuit.reference_sample"></a>
+```python
+# stim.Circuit.reference_sample
+
+# (in class stim.Circuit)
+def reference_sample(
+    self,
+    bit_packed: bool = False,
+) -> np.ndarray:
+    """Samples the given circuit in a deterministic fashion.
+
+    Discards all noisy operations, and biases all collapse events
+    towards +Z instead of randomly +Z/-Z.
+
+    Args:
+        circuit: The circuit to "sample" from.
+        bit_packed: Defaults to False. Determines whether the output numpy arrays
+            use dtype=bool_ or dtype=uint8 with 8 bools packed into each byte.
+
+    Returns:
+        reference_sample: reference sample sampled from the given circuit.
     """
 ```
 
@@ -3391,6 +3429,7 @@ def __init__(
     *,
     skip_reference_sample: bool = False,
     seed: object = None,
+    reference_sample: object = None,
 ) -> None:
     """Creates a measurement sampler for the given circuit.
 
@@ -3439,6 +3478,15 @@ def __init__(
             CAUTION: simulation results *MAY NOT* be consistent if you vary how many
             shots are taken. For example, taking 10 shots and then 90 shots will
             give different results from taking 100 shots in one call.
+        reference_sample: The data to xor into the measurement flips produced by the
+            frame simulator, in order to produce proper measurement results.
+            This can either be specified as an `np.bool_` array or a bit packed
+            `np.uint8` array (little endian). Under normal conditions, the reference
+            sample should be a valid noiseless sample of the circuit, such as the
+            one returned by `circuit.reference_sample()`. If this argument is not
+            provided, the reference sample will be set to
+            `circuit.reference_sample()`, unless `skip_reference_sample=True`
+            is used, in which case it will be set to all-zeros.
 
     Returns:
         An initialized stim.CompiledMeasurementSampler.
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -1040,8 +1040,8 @@ def compile_sampler(
     self,
     *,
     skip_reference_sample: bool = False,
-    seed: object = None,
-    reference_sample: object = None,
+    seed: Optional[int] = None,
+    reference_sample: Optional[np.ndarray] = None,
 ) -> stim.CompiledMeasurementSampler:
     """Returns an object that can quickly batch sample measurements from the circuit.
 
@@ -2003,6 +2003,7 @@ def num_ticks(
 # (in class stim.Circuit)
 def reference_sample(
     self,
+    *,
     bit_packed: bool = False,
 ) -> np.ndarray:
     """Samples the given circuit in a deterministic fashion.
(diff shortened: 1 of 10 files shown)


src/stim/circuit/circuit.pybind.cc
```diff
@@ -458,6 +459,12 @@ void stim_pybind::pybind_circuit_methods(pybind11::module &, pybind11::class_<Ci
                     CAUTION: simulation results *MAY NOT* be consistent if you vary how many
                     shots are taken. For example, taking 10 shots and then 90 shots will
                     give different results from taking 100 shots in one call.
+                reference_sample: Reference sample used by the sampler. If None
```
Strilanc, line 462: Describe in more detail what the reference sample is. Most people reading this documentation will not know what it is.

```
    reference_sample: The data to xor into the measurement flips produced by the frame
        simulator, in order to produce proper measurement results. This can either be specified
        as an `np.bool_` array or a bit packed `np.uint8` array (little endian). Under normal
        conditions, the reference sample should be a valid noiseless sample of the circuit, such as
        the one returned by `circuit.reference_sample()`. If this argument is not provided, the
        reference sample will be set to `circuit.reference_sample()`, unless
        `skip_reference_sample=True` is used, in which case it will be set to all-zeros.
```

src/stim/py/compiled_measurement_sampler.pybind.cc
```diff
@@ -66,10 +66,22 @@ pybind11::class_<CompiledMeasurementSampler> stim_pybind::pybind_compiled_measur
...
-    const Circuit &circuit, bool skip_reference_sample, const pybind11::object &seed) {
-    simd_bits<MAX_BITWORD_WIDTH> ref_sample = skip_reference_sample
-                                                  ? simd_bits<MAX_BITWORD_WIDTH>(circuit.count_measurements())
-                                                  : TableauSimulator::reference_sample_circuit(circuit);
+    const Circuit &circuit,
+    bool skip_reference_sample,
+    const pybind11::object &seed,
+    const pybind11::object &reference_sample) {
+    uint64_t num_bits = circuit.count_measurements();
+    simd_bits<MAX_BITWORD_WIDTH> ref_sample(circuit.count_measurements());
```
Strilanc, line 74: Can you avoid unnecessarily initialization and allocating a reference sample that will be replaced in most cases? This thing might be megabytes big; it's a large cost to pay to unconditionally initialize it like this.

src/stim/py/compiled_measurement_sampler.pybind.cc
```diff
@@ -129,6 +142,9 @@ void stim_pybind::pybind_compiled_measurement_sampler_methods(
                     CAUTION: simulation results *MAY NOT* be consistent if you vary how many
                     shots are taken. For example, taking 10 shots and then 90 shots will
                     give different results from taking 100 shots in one call.
+                reference_sample: Reference sample used by the sampler. If None
```
Strilanc, line 145: copy paste better desc from above

src/stim/py/compiled_measurement_sampler_pybind_test.py
```diff
@@ -118,6 +119,37 @@ def test_skip_reference_sample():
...
+        circuit.compile_sampler(reference_sample=ref_sample).sample(1),
+        [[True]],
+    )
+    ref_sample = s.reference_sample(circuit)
+    np.testing.assert_array_equal(
+        circuit.compile_sampler(reference_sample=ref_sample).sample(1),
+        [[True]],
+    )
+    with pytest.raises(ValueError):
+        circuit.compile_sampler(reference_sample=ref_sample, skip_reference_sample=True).sample(1)
```
Strilanc, line 151: drop the `.sample(1)` since it's past the exception

src/stim/py/compiled_measurement_sampler_pybind_test.py
```diff
@@ -118,6 +119,37 @@ def test_skip_reference_sample():
...
+    )
+    circuit = stim.Circuit("X 0\nM 0")
+    s = stim.TableauSimulator()
+    ref_sample = np.array([False])
+    np.testing.assert_array_equal(
+        circuit.compile_sampler(reference_sample=ref_sample).sample(1),
+        [[False]],
+    )
+    ref_sample = s.reference_sample(circuit)
+    np.testing.assert_array_equal(
```
Strilanc, line 135: Test that specifying a bit packed reference sample works. Make sure you have at least 9 measurements so it crosses byte boundaries. It should be the case that the 1s bit is the first measurement, the 2s bit the second, the 4s bit the fourth, etc.

src/stim/py/compiled_measurement_sampler_pybind_test.py
```diff
@@ -118,6 +119,37 @@ def test_skip_reference_sample():
         [[True]],
     )
 
+def test_reference_sample_init():
+    np.testing.assert_array_equal(
+        stim.Circuit("X 0\nM 0").compile_sampler(reference_sample=None).sample(1),
+        [[True]],
+    )
+    circuit = stim.Circuit("X 0\nM 0")
```
Strilanc, line 127: Add a sanity check that specifying `circuit.reference_sample()` gives the same samples as specifying None, for a circuit with deterministic measurements.

src/stim/py/compiled_measurement_sampler_pybind_test.py
```diff
@@ -118,6 +119,37 @@ def test_skip_reference_sample():
         [[True]],
     )
 
+def test_reference_sample_init():
+    np.testing.assert_array_equal(
```
Strilanc, line 123: Test a mix of commuting and anticommuting and redundant measurements, such as `MPP X1*X2 Y1*Y2 Z1*Z2`.

Note you can use `seed=0` to force the same underlying samples between a no-ref-sample and yes-ref-sample sampler to be identical, allowing you to check the ref is used correctly. Note seeds change meaning from version to version so you need to do the comparison, not just hardcode.

src/stim/simulators/tableau_simulator.pybind.cc
```diff
@@ -2019,4 +2020,31 @@ void stim_pybind::pybind_tableau_simulator_methods(pybind11::module &m, pybind11
                 )
         )DOC")
             .data());
+    c.def(
+        "reference_sample",
```
Strilanc, line 2024: I don't think this method makes sense on the `stim.TableauSimulator` class. The issue is that it is very stateful, for example the user may have applied Hadamard gates. What does it mean to ask for a reference sample from the last half of a circuit? Its contents can depend very sensitively on the beginning of the circuit.

Instead of adding it here, add it as a method on `stim.Circuit`.

Strilanc (changes requested):
Nice change. Lots of comments to tweak it; all superficial except moving the method to a different class.

src/stim/py/compiled_measurement_sampler.pybind.cc
fdmalone: I moved the initialization / returns into the if/else if that works?

Strilanc: Yeah that should work.

doc/python_api_reference_vDev.md
```diff
@@ -1040,6 +1041,7 @@ def compile_sampler(
     *,
     skip_reference_sample: bool = False,
     seed: object = None,
+    reference_sample: object = None,
```
Strilanc, line 1044: Make this be `np.ndarray` instead of `object` by adding an `@signature` line to the docstring, similar to how it's done in other methods.

doc/python_api_reference_vDev.md
```diff
@@ -3391,6 +3429,7 @@ def __init__(
...
```
Strilanc, line 3432: `@signature` for `np.ndarray`

src/stim/circuit/circuit.pybind.cc
```diff
@@ -471,6 +486,33 @@ void stim_pybind::pybind_circuit_methods(pybind11::module &, pybind11::class_<Ci
...
+        [](const Circuit &self, bool bit_packed) {
+            simd_bits<MAX_BITWORD_WIDTH> ref = TableauSimulator::reference_sample_circuit(self);
+            simd_bits_range_ref<MAX_BITWORD_WIDTH> reference_sample(ref.ptr_simd, ref.num_simd_words);
+            size_t num_measure = self.count_measurements();
+            return simd_bits_to_numpy(reference_sample, num_measure, bit_packed);
+        },
+        pybind11::kw_only(),
+        pybind11::arg("bit_packed") = false,
+        clean_doc_string(R"DOC(
+            @signature def reference_sample(self, bit_packed: bool = False) -> np.ndarray:
```
Strilanc, line 500: Add the `, *,` before `bit_packed` indicating the `kw_only`

src/stim/circuit/circuit_pybind_test.py
```diff
@@ -1345,3 +1345,32 @@ def test_num_ticks():
...
+    assert len(ref) == 0
+    circuit = stim.Circuit(
+        """
+        H 0
+        M 0
+        M 1
+    """
+    )
+    ref = circuit.reference_sample()
+    assert ref[0] == ref[1]
```
Strilanc, line 1367: Add `np.testing.assert_array_equal(circuit.reference_sample(), circuit.reference_sample())` to check that it's consistent

Strilanc, line 1367: I am hesitant to guarantee what the *specific* reference sample is, across versions. I think a better test might be...

```
H 0 1
CX 0 2 1 3
MPP X0*X1 Y0*Y1 Z0*Z1
```

Assert that there's an odd number of non-zero results.

src/stim/circuit/circuit_pybind_test.py
```diff
@@ -1345,3 +1345,32 @@ def test_num_ticks():
...
+    circuit = stim.Circuit()
+    circuit.append("X", (i for i in range(0, 100, 2)))
+    circuit.append("M", (i for i in range(100)))
+    ref_sample = circuit.reference_sample(bit_packed=False)
+    expected = np.zeros((len(ref_sample) // 8 + 1), dtype=np.uint8)
+    for bit_indx, bit in enumerate(ref_sample):
+        expected[bit_indx//8] |= (bit << (bit_indx % 8))
```
Strilanc, line 1374: Use `np.unpackbits` (little endian)  and `np.testing.assert_array_equal` for the comparison, instead of doing it custom here.

Strilanc (changes requested):
Very close! Just some doc and test tweaks left.

doc/python_api_reference_vDev.md
fdmalone: When I add a signature and regenerate the docs it doesn't seem to update? Am I missing something simple.

Strilanc: Hm, maybe make sure it's the very first line of the docstring? Try copy pasting from another place using `@signature` and confirm it shows that one, so there's as few differences as possible.

This is all just some jank search and replace stuff done in `glue/python/generate_stub_file.py`

fdmalone: ugh, I forgot to rebuild the package so naturally import stim.Circuit etc was out of date lol

Strilanc (approved):
LGTM

PR #556, opened 2023-05-21
Remove hardcoded SIMD width from `stim/io` directory

markturner289:
This PR is the first in a set of PRs that aim to unblock #548. It focusses purely on I/O components in Stim and will contribute to the unblocking of #548 by removing the dependency of all I/O components on `MAX_BITWORD_WIDTH`. It does this via further templatisation.

Specifically, the proposed templatisation encompasses:
 - The following class templates:
   - `MeasureRecordBatch<W>`
   - `MeasureRecordReader<W>` (and all derived classes)
 - The following function templates:
   - `MeasureRecordBatchWriter::batch_write_bit<W>`
   - `MeasureRecordBatchWriter::batch_write_byte<W>`
   - `stim::transposed_vs_ref<W>`
   - `stim::write_table_data<W>`

While this PR's primary motivation is to facilitate templatising simulators, this PR has also implemented testing of I/O components across all SIMD widths via the `TEST_EACH_WORD_SIZE_W` macro.


Strilanc (approved):
This all looks good. Really appreciate this getting done; looks like the expected slog.

PR #558, opened 2023-05-25
Remove hardcoded SIMD width from `stim/circuit` directory

markturner289:
This PR removes the dependency of all Circuit components on `MAX_BITWORD_WIDTH`. It does this via templatisation of `decompose_mpp_operation` (the only place where `MAX_BITWORD_WIDTH` is used in `src/circuit` outside the benchmarks, pybind etc). I had to do some light splitting-up of files (notably `circuit.h`) to facilitate this.

As before, happy to make any changes.

### DIFF
--- a/file_lists/source_files_no_main
+++ b/file_lists/source_files_no_main
@@ -1,6 +1,7 @@
 src/stim.cc
 src/stim/arg_parse.cc
 src/stim/circuit/circuit.cc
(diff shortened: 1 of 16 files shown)


Strilanc (approved):
LGTM. I like how it splits out CircuitInstruction and adds `[[fallthrough]]`.

PR #559, opened 2023-06-05
Remove hardcoded SIMD width from `stim/stabilizer` dir

markturner289:
This PR removes the dependency of all stabilizer components on `MAX_BITWORD_WIDTH`. It does this via templatisation. This primarily aims to facilitate templatised simulators (#548), but also adds extensive testing of stabilizer components across all SIMD widths via the `TEST_EACH_WORD_SIZE_W` macro.

The following class templates are introduced:
 - `PauliString<W>`
 - `PauliStringRef<W>`
 - `Tableau<W>`
 - `TableauHalf<W>`
 - `TableauTransposedRaii<W>`
 - `CommutingPauliStringIterator<W>`
 - `TableauIterator<W>`
 - `VectorSimulator<W>` (ideally no simulators would have been templatised in this PR, but some components in `stim/stabilizers` depend on this simulator, making this necessary) [EDIT: NO LONGER THE CASE!]

The following function templates are introduced:
 - Templatised ostream operators for the above
 - `std::random_stabilizer_tableau_raw<W>`
 - `stim::Gate::tableau<W>`
 - `stabilizers_to_tableau<W>`
 - `unitary_to_tableau<W>`
 - `tableau_to_circuit<W>`
 - `circuit_to_output_state_vector<W>`
 - `circuit_to_tableau<W>`
 - `stabilizer_state_vector_to_circuit<W>`
 - `tableau_to_unitary<W>`

The `TableauSimulator` has not yet been templatised and this blocks a minority of tests in `stim/stabilizer` from being tested with `TEST_EACH_WORD_SIZE_W`. I have left comments below, highlighting these cases.


src/stim/circuit/gate_data.test.cc
```diff
@@ -67,7 +68,8 @@ TEST(gate_data, hash_matches_storage_location) {
     }
 }
 
-std::pair<std::vector<PauliString>, std::vector<PauliString>> circuit_output_eq_val(const Circuit &circuit) {
+std::pair<std::vector<PauliString<MAX_BITWORD_WIDTH>>, std::vector<PauliString<MAX_BITWORD_WIDTH>>>
+circuit_output_eq_val(const Circuit &circuit) {
```
markturner289, line 72: The dependency on the `TableauSimulator` means that tests using this function cannot be tested for all SIMD widths until the `TableauSimulator` is templatised. The only test this impacts here is `decompositions_are_correct`.

src/stim/simulators/vector_simulator.inl
```diff
@@ -21,14 +21,15 @@
 #include "stim/probability_util.h"
 #include "stim/stabilizers/pauli_string.h"
 
-using namespace stim;
+namespace stim {
 
-VectorSimulator::VectorSimulator(size_t num_qubits) {
+template <size_t W>
+VectorSimulator<W>::VectorSimulator(size_t num_qubits) {
```
Strilanc, line 27: Wait, why is the vector simulator getting a word width? It doesn't store any simd bit fields. Shouldn't only some of its methods require this parameter?

Strilanc (changes requested):
You can really tell that PauliString was the big one to change as it gets basically everywhere in the code base.

Main thing I'm wondering about is why VectorSimulator became VectorSimulator<W>, instead of only some of its methods getting templatized.

src/stim/simulators/vector_simulator.inl
markturner289: Probably an oversight on my part.

You're right, it has no member data that depends on W. I must have templatised it because of the use of the templatised class `PauliStringRef`.

I'll change this so that templatisation only touches the 4 member functions in the VectorSimulator and not the class as a whole.

Generally this was the pattern I have been trying to following - templatise functions unless member data also needs the template parameter.

Do you foresee any utility in having the class template here in future or shall I make this change now?

Strilanc: Make the change back to not-templated-class. VectorSimulator works with floats not bits so it's not relevant to it.

markturner289: Change made.

Strilanc:
Will be merged by https://github.com/quantumlib/Stim/pull/572 , where I fixed the massive merge conflicts I accidentally created

PR #565, opened 2023-06-07
Add iframes around HTML widgets to isolate them in Jupyter notebooks

Strilanc:


src/stim/cmd/command_diagram.pybind.cc
```diff
@@ -44,8 +44,25 @@ pybind11::class_<DiagramHelper> stim_pybind::pybind_diagram(pybind11::module &m)
     return c;
 }
 
+// shamelessly stolen from https://stackoverflow.com/a/9907752
+std::string escape(const std::string& src) {
```
Strilanc, line 48: escape what? Add a better name.

src/stim/cmd/command_diagram.pybind.cc
```diff
@@ -55,17 +72,23 @@ void stim_pybind::pybind_diagram_methods(pybind11::module &m, pybind11::class_<D
...
         }
         if (self.type == DIAGRAM_TYPE_HTML) {
-            return pybind11::cast(self.content);
+            output = self.content;
+        }
+	if (output == "None"){
+	    return pybind11::none();
+        } else {
+            std::string prefix = "<iframe width=950 height=650 srcdoc=\"";
+            std::string postfix = "\"></iframe>";
```
Strilanc, line 89: If you refactor the `escape` method into a `wrap html in iframe` method, you can avoid the need for the `"None"` sygil by just wrapping inplace. I'd actually prefer that, since it's not obvious to me that we unconditionally want to wrap all html.

Strilanc:
Obsoleted by https://github.com/quantumlib/Stim/pull/608

PR #594, opened 2023-07-18
New error messages in search algos

### LINKED ISSUE #593: Save users from themselves: shortest_graphlike_error
`shortest_graphlike_error` should produce helpful error messages in specific, common, stupid cases:

1. the circuit has no observables
2. the circuit has no error instructions
3. the circuit has no detectors

I could take or leave the last one, because hypothetically someone might find a use for say inserting only a few errors in a detectorless circuit and trying to find if they ever touched an observable?

Probably should do this for `find_undetectable_logical_error` as well

mmcewen-g:
fixes #593

### DIFF
--- a/src/stim/circuit/circuit_pybind_test.py
+++ b/src/stim/circuit/circuit_pybind_test.py
@@ -716,34 +716,34 @@ def test_shortest_graphlike_error_empty():
 def test_shortest_graphlike_error_msgs():
     with pytest.raises(
             ValueError,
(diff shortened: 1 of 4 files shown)


Strilanc (changes requested):
Nice change.

This also needs a `stim.DetectorErrorModel.shortest_graphlike_error` test, and a test of the non-graphlike error search.

Strilanc (approved):

Issue #397, opened 2022-10-28
Add `stim.PauliString.iter_all`


Strilanc:
This method should return an iterator that iterates through all pauli strings matching the filters specified by the arguments given to the method. The signature should be:

```
@staticmethod
def iter_all(
    length: int,
    *,
    min_weight: int = 0,
    max_weight: Optional[int] = None,
    allowed_paulis: str = "XYZ",
) -> Iterator[stim.PauliString]:
    ...
```

Example:

```
>>> for e in stim.PauliString.iter_all(3, min_weight=1, max_weight=2, allowed_paulis="XZ"):
...    print(e)
+X__
+Z__
+_X_
+_Z_
+__X
+__Z
+XX_
+XZ_
+ZX_
+ZZ_
+X_X
+X_Z
+Z_X
+Z_Z
+_XX
+_XZ
+_ZX
+_ZZ
```

It's conceivable that some sort of `out: Optional[stim.PauliString] = None` argument could also be included, to give the underlying C++ code a place to put the results. This would avoid having to allocate a new object for every result, which I expect to be the most expensive part of the iteration. But maybe benchmarking allocate-results vs not-allocate-results would show it's swamped by other python overheads and therefore not worth doing.

`length` determines the output length of the pauli strings. There are 4^length potential unsigned outputs.

The `min_weight` and `max_weight` arguments control how many non-identity terms there can be. Ideally, the method would iterate terms in order of increasing weight. max weight defaults to `length` if not set.

The `allowed_paulis` argument determines what paulis (X, Y, and/or Z) are allowed to appear in the iterated strings. If it's set to a single pauli, this problem becomes equivalent to iterating bit strings matching the requested weights. If it's two paulis, it's iterating satisfying bit strings m and iterating partitions of those bit strings (x ^ z == m). If it's three then it's unions instead of partitions (x | z == m).

It's important that the time scaling should be proportional to the number of outputs, not to the number of pauli strings of a given length. Iterating weight 3 and weight 4 strings should take O(n^3 + n^4) time, not O(4^n) time.

fdmalone:
Is the idea here it to produce all 4^N  possible Pauli strings (or a subset of them dependent on the weight) given N qubits (times 4 for +/-1 and +/- i)? Sounds like a small modification of [this](https://graphics.stanford.edu/~seander/bithacks.html#NextBitPermutation) on the C++ side or some itertools gymnastics on the python side.

Strilanc:
[...]

PR #598, opened 2023-07-21
Pauli iter all

fdmalone:
Adds stim.PauliString.iter_all method.

Following the discussion in #397 the algorithm is a combination of finding the next lexicographically ordered permutation of w bits followed by iterating over all 3^w PauliStrings given this permutation of the qubit labels. For the first part I modified the bit twiddle algorithm to account for multiple words, for the second part I just iterate over 3^w integers and map this to a Pauli using the ternary representation of the integer. This was a bit trickier to get right than I expected, plenty of edge cases.

The modifications for the bit twiddle algorithm are quite specific and I was a bit torn between adding more general operators to simd_bits (like left / right shift + subtraction) which would be cleaner, and what I did, which is a bit clunky.

There are also a few optimizations that could be made which I haven't:

1. for small w I could avoid the repeated loop over 3^w and store the Pauli strings in the first pass, rather than repeating this work for each of the num_qubits C w permutations (combinations))
2. some bitwise algorithms are suboptimal (counting trailing zeros) and/or may already exist in stim but I couldn't find them. Happy to change these. 
3. The decoded bit labels / locations could be stored rather than decode each time.

Draft for the moment until:
- [x] Fix pybind issue.
- [ ] Profile against naive python implmentation.
- [x] Check docstrings / algorithm description


doc/python_api_reference_vDev.md
```diff
@@ -7793,6 +7837,93 @@ def to_unitary_matrix(
...
+
+# (in class stim.PauliStringIterator)
+def next_qubit_permutation(
+    self,
+) -> object:
+    """Get the next permutation of qubit labels.
+
+    It's alot easier to test more complicated edge cases in python
+    which largely arise due to the algorithm for generating the next
+    permutation. The user should not interact with this function.
```
Strilanc: If the user shouldn't interact with the function, start its name with something like `__internal_` and throw in a `@deprecated` in the docstring so it doesn't show in the api reference.

src/stim/py/numpy.pybind.h
```diff
@@ -31,6 +31,9 @@ pybind11::object simd_bits_to_numpy(
     stim::simd_bits_range_ref<stim::MAX_BITWORD_WIDTH> bits, size_t num_bits, bool bit_packed);
 void memcpy_bits_from_numpy_to_simd(
     size_t num_bits, const pybind11::object &src, stim::simd_bits_range_ref<stim::MAX_BITWORD_WIDTH> dst);
+size_t numpy_to_size(const pybind11::object &numpy_array, size_t expected_size);
```
Strilanc, line 34: Since these are moving into a header, make the name longer and clearer like `bit_count_of_bit_packable_numpy_array` or something.

src/stim/stabilizers/pauli_string.pybind.cc
```diff
@@ -636,6 +585,54 @@ void stim_pybind::pybind_pauli_string_methods(pybind11::module &m, pybind11::cla
...
+                _min_weight = 0;
+            } else {
+                _min_weight = pybind11::cast<size_t>(min_weight);
+            }
+            return PauliStringIterator<stim::MAX_BITWORD_WIDTH>(num_qubits, _min_weight, _max_weight);
+        },
+        pybind11::arg("num_qubits"),
+        pybind11::kw_only(),
+        pybind11::arg("min_weight") = pybind11::none(),
+        pybind11::arg("max_weight") = pybind11::none(),
```
Strilanc, line 607: A useful argument here would be `allowed_paulis: str`, so the user could restrict to X errors (allowed_paulis="X") or not-Y-errors (allowed_paulis="XZ").

src/stim/stabilizers/pauli_string.pybind.cc
```diff
@@ -636,6 +585,54 @@ void stim_pybind::pybind_pauli_string_methods(pybind11::module &m, pybind11::cla
...
+                max_weight: The maximum weight PauliString to consider. If None
+                    then max_weight is set to num_qubits.
+
+            Returns:
+                An Iterable[stim.PauliString] that yields the requested PauliStrings.
+
+            Examples:
+                >>> import stim
+                >>> n = 0
+                >>> for t in stim.PauliString.iter_all(4, 2, 3):
```
Strilanc: missing keywords on args

src/stim/stabilizers/pauli_string_iter.inl
```diff
@@ -0,0 +1,306 @@
...
+            // 10 = YX
+            // 11 = YZ
+            // 12 = YY  ...
+            // This inner while loop will yield r and indx, where r is the
+            // trit (0, 1, 2) and indx is the corresponding qubit index.
+            int r = n % 3;
+            n /= 3;
+            if (r == 2) {
+                // Already have an X set so just set the Z.
+                result.zs.u64[set_bits[indx] / 64] ^= uint64_t{1} << (set_bits[indx] & 63);
```
Strilanc, line 196: `result.zs[indx] = b` will work

src/stim/stabilizers/pauli_string_iter.inl
```diff
@@ -0,0 +1,306 @@
...
+            // Recall the ternary ordering for cur_w = 2, num_qubits = 2 would yield
+            // 00 = XX
+            // 01 = XZ
+            // 02 = XY
+            // 10 = YX
+            // 11 = YZ
+            // 12 = YY  ...
+            // This inner while loop will yield r and indx, where r is the
+            // trit (0, 1, 2) and indx is the corresponding qubit index.
+            int r = n % 3;
```
Strilanc, line 192: If you use adjacent bits in a word to encode a pauli (00=I, 01=X, ...) then you can use `a++` followed by `a |= ~(a | (a >> 1)) & 0x5555555555555555ULL` to skip over values that would have identity paulis.

```
def bits_to_paulis(x: int, n: int) -> str:
    bits = bin(x)[2:].rjust(n*2, '0')[-n*2:]
    return ''.join('_XYZ'[int(bits[k:k+2], 2)] for k in range(0, len(bits), 2))


a = 0
for _ in range(3**3):
    a += 1
    a |= ~(a | (a >> 1)) & 0x5555555555555555
    print(bits_to_paulis(a, 4))
```

prints:

```
XXXX
XXXY
XXXZ
XXYX
XXYY
XXYZ
XXZX
XXZY
XXZZ
XYXX
XYXY
XYXZ
XYYX
XYYY
XYYZ
XYZX
XYZY
XYZZ
XZXX
XZXY
XZXZ
XZYX
XZYY
XZYZ
XZZX
XZZY
XZZZ
```

src/stim/stabilizers/pauli_string_iter.perf.cc
```diff
@@ -0,0 +1,35 @@
...
+BENCHMARK(PauliStringIter) {
+    size_t num_qubits = 128;
+    size_t min_weight = 1;
+    size_t max_weight = 2;
+    PauliStringIterator<MAX_BITWORD_WIDTH> iter(num_qubits, min_weight, max_weight);
+    size_t c = 0;
+    benchmark_go([&]() {
+        while (iter.iter_next()) {
+            c += iter.result.num_qubits;
+            std::cout << iter.result << std::endl;
```
Strilanc, line 30: This block will be run a variable number of times. Don't write to stdout as part of it, and recreate the iterator since otherwise it will exhaust after the first time.

doc/python_api_reference_vDev.md
```diff
@@ -7793,6 +7838,101 @@ def to_unitary_matrix(
...
+<a name="stim.PauliStringIterator.seed_iterator"></a>
+```python
+# stim.PauliStringIterator.seed_iterator
+
+# (in class stim.PauliStringIterator)
+def seed_iterator(
+    self,
+    arg0: object,
+) -> None:
+    """Seed the iterator with a given qubit pattern.
```
Strilanc, line 7923: Not clear to me what this means.

Strilanc (commented):
Looking good so far. A few ideas commented.

fdmalone:
Sorry, I forced pushed when it was still a draft to clean up the many doc related fixes. Should be ready for review now sans profiling.

Strilanc:
> I was a bit torn between adding more general operators to simd_bits (like left / right shift + subtraction) which would be cleaner, and what I did, which is a bit clunky

I'd be fine with those additions to `simd_bits`.

Strilanc:
```
from typing import Iterable


def bits_to_paulis(x: int, n: int) -> str:
    bits = bin(x)[2:].rjust(n*2, '0')[-n*2:]
    return ''.join('_XYZ'[int(bits[k:k+2], 2)] for k in range(0, len(bits), 2))


def count_trailing_zeros(x: int) -> int:
    t = 0
    while not (x & 1):
        x >>= 1
        t += 1
    return t


def masked_increment(x: int, mask: int) -> int:
    return ((x | ~mask) + 1) & mask


def next_bitstring_of_same_hamming_weight(x: int) -> int:
    c1 = x | (x - 1)
    c2 = c1 + 1
    c3 = (~c1 & -~c1) - 1
    c4 = c3 >> (count_trailing_zeros(x) + 1)
    return c2 | c4


def iter_pauli_strings(weight: int, length: int) -> Iterable[str]:
    hamming_mask = (1 << weight) - 1
    while hamming_mask < 2**length:
        # Spread out bits into pairs.
        h = hamming_mask
        mh = 0b0000000000000000000000000000000011111111111111110000000000000000
        ml = 0b0000000000000000000000000000000000000000000000001111111111111111
        h = (h & ml) | ((h & mh) << 16)
        mh = 0b0000000000000000111111110000000000000000000000001111111100000000
        ml = 0b0000000000000000000000001111111100000000000000000000000011111111
        h = (h & ml) | ((h & mh) << 8)
        mh = 0b0000000011110000000000001111000000000000111100000000000011110000
        ml = 0b0000000000001111000000000000111100000000000011110000000000001111
        h = (h & ml) | ((h & mh) << 4)
        mh = 0b0000110000001100000011000000110000001100000011000000110000001100
        ml = 0b0000001100000011000000110000001100000011000000110000001100000011
        h = (h & ml) | ((h & mh) << 2)
        mh = 0b0010001000100010001000100010001000100010001000100010001000100010
        ml = 0b0001000100010001000100010001000100010001000100010001000100010001
        h = (h & ml) | ((h & mh) << 1)
        h |= h << 1

        # Iterate over non-00 values of masked bit pairs, with 00 elsewhere
        xz_mask = 0
        for _ in range(3**weight):
            xz_mask = masked_increment(xz_mask, h)
            xz_mask |= ~(xz_mask | (xz_mask >> 1)) & 0b0101010101010101010101010101010101010101010101010101010101010101
            xz_mask &= h
            yield bits_to_paulis(xz_mask, length)

        # Next mask
        hamming_mask = next_bitstring_of_same_hamming_weight(hamming_mask)


t = 0
for e in iter_pauli_strings(weight=3, length=20):
    print(e)
    t += 1
print(t)
```

Strilanc:
A better one (in particular see `pair_sat_increment`):

```
import math
from typing import Iterable, Tuple


def bits_to_paulis(x: int, z: int, n: int) -> str:
    s = ''
    for k in range(n):
        s += '_XYZ'[(x & 1) + (z & 1) * 2]
        x >>= 1
        z >>= 1
    return s[::-1]


def count_trailing_zeros(x: int) -> int:
    t = 0
    while not (x & 1):
        x >>= 1
        t += 1
    return t


def masked_increment(x: int, mask: int) -> int:
    return ((x | ~mask) + 1) & mask


def next_bitstring_of_same_hamming_weight(x: int) -> int:
    c1 = x | (x - 1)
    c3 = ((c1 + 1) & ~c1) - 1
    c4 = c3 >> (count_trailing_zeros(x) + 1)
    return (c1 + 1) | c4


def pair_sat_increment(x: int, z: int, m: int) -> Tuple[int, int]:
    """Finds the next (x, z) such that x | z == m."""
    inc = x & z
    up = ~inc
    inc |= ~m
    inc += 1
    inc &= m
    up &= inc
    z &= inc | ~x
    z ^= x & up
    x ^= up
    return x, z


def iter_pauli_strings(weight: int, length: int) -> Iterable[str]:
    h = (1 << weight) - 1
    while h < 2**length:
        x, z = h, 0
        for _ in range(3**weight):
            yield bits_to_paulis(x, z, length)
            x, z = pair_sat_increment(x, z, h)
        h = next_bitstring_of_same_hamming_weight(h)


t = 0
seen = set()
for e in iter_pauli_strings(weight=4, length=10):
    assert e not in seen, e
    seen.add(e)
    print(e)
    t += 1
print(t, 3**4 * math.factorial(10) // math.factorial(6) // math.factorial(4))
```

src/stim/stabilizers/pauli_string_iter.inl
fdmalone: I'll replace this function with one of your suggestions.

doc/python_api_reference_vDev.md
fdmalone: I'm actually going to remove this function, after closer thought it's not super helpful. The idea was to be able to "seed" the iterator at a specific bit pattern which may be difficult to reach if w was large, but this can be tested on the C++ side more easily, and not sure if it's actually useful on the python side.

fdmalone: Actually scratch that, I needed it for testing random starting points for long strings with higher weight. I will change the name to something more descriptive

fdmalone:
Ok, I can replace my ternary iteration with the bit twiddle above.

src/stim/stabilizers/pauli_string.pybind.cc
fdmalone: Good point.

fdmalone:
In particular I will add additional ops to simd_bits (left/right shift and an adder) and replace my ternary iteration with the bit twiddle. I may separate out the new operations in a separate PR.

Strilanc:
SGTM

src/stim/stabilizers/pauli_string_iter.inl
fdmalone: resolved with pair_sat_increment.

doc/python_api_reference_vDev.md
fdmalone: I renamed this to set_current_permutation.

fdmalone:
The windows failures seem to be due to a memory error

``` python
          pauli_it = list(
              stim.PauliString.iter_all(
  >               num_qubits, min_weight=min_weight, max_weight=max_weight
              )
          )
  E       MemoryError
```

The test_iter_all_random_permutation tests also seems to trigger test failures on windows periodically on win32 platforms. Not sure what's going on there.

src/stim/stabilizers/pauli_string.pybind.cc
fdmalone: I still need to add this, and also the random selection of signs/phases.

src/stim/stabilizers/pauli_string_iter.inl
```diff
@@ -0,0 +1,296 @@
...
+        // z &= inc | ~x
+        simd_bits<W> comp_x(result.xs);
+        comp_x.invert_bits();
+        result.zs &= inc | comp_x;
+        // z ^= x & up
+        result.zs ^= result.xs & up;
+        // x ^= up
+        result.xs ^= up;
+        cur_k++;
+        return true;
```
Strilanc, line 190: Why is this a `while` loop if it always exits?

src/stim/stabilizers/pauli_string_iter.inl
```diff
@@ -0,0 +1,296 @@
...
+    // next_zero - num_zeros - 1 - words * 64 is always < 64, so don't need to worry about overflow like above.
+    t3.u64[word_loc] = (1ULL << (next_zero - num_zeros - 1 - word_loc * 64)) - 1;
+    cur_perm = t1;
+    cur_perm ^= t2;
+    cur_perm |= t3;
+}
+
+template <size_t W>
+bool PauliStringIterator<W>::pair_sat_increment() {
+    // This will overflow for large cur_w.
```
Strilanc, line 154: Overflow is bad or good here?

src/stim/stabilizers/pauli_string_iter.inl
```diff
@@ -0,0 +1,296 @@
...
+    size_t num_terms = static_cast<size_t>(pow(3, cur_w));
```
Strilanc, line 155: Are you sure this actually returns the right answer for all relevant values? Doubles have 53 bits of precision; not enough for a 64 bit integer. It'd be safer to have a method like

```
uint64_t pow3(uint64_t p) {
  assert(p < 41); 
  uint64_t r = 1;
  if (p & 1) r *= 3
  if (p & 2) r *= 9;
  if (p & 4) r *= 81;
  if (p & 8) r *= 6561;
  if (p & 16) r *= 43046721;
  if (p & 32) r *= 1853020188851841;
  return r;
}
```

src/stim/stabilizers/pauli_string_iter.inl
```diff
@@ -0,0 +1,296 @@
...
+        simd_bits<W> up(inc);
+        up.invert_bits();
+        // inc |= ~m
+        simd_bits<W> comp_m(cur_perm);
+        comp_m.invert_bits();
+        inc |= comp_m;
+        simd_bits<W> one(cur_perm.num_bits_padded());
+        one.u64[0] = uint64_t{1};
+        // inc += 1
+        inc += one;
```
Strilanc, line 176: Maybe add a `++x` method to `simd_bits`?

Strilanc (changes requested):
I think the major tasks left are debugging the windows crash and adding pytest unit tests using the python side of the API.

src/stim/stabilizers/pauli_string_iter.inl
fdmalone: Yeah, it probably makes sense to add -- / -= too. For example, there are a few awkward parts where I'm doing (1 << (sum_number_greather_than_64) - 1.

fdmalone:
I'm going to close this as I won't have time to come back to it for another month or so.

PR #600, opened 2023-08-02
Add subcommand to convert to/from various result formats

justinledford:
This can be very useful when comparing results from experiments when the outputs were stored in different formats.


Strilanc (changes requested):
This is a good start, but there are several more use cases that I think need to be covered. Extend this to include the following use cases:

- I have the DEM, not the circuit, and want to convert detection event data.
- I have neither the DEM nor the circuit but I know all the values that would be derived from them like the bits per shot
- I know how many bits per shot, but I don't know if it's detection event data or measurement data (and I'm not using the DETS format which is the only one where that distinction matters)
- The input file has observable data appended to the detector data, and I want to split the observable data out into a separate file

These use cases all kind of differ in what information they need. In general the command should infer everything it can from whatever the user provides, and if it can't infer a needed quantity it should raise an error identifying that quantity and saying various ways it can be provided.

src/stim/cmd/command_convert.cc
```diff
@@ -0,0 +1,261 @@
...
+    FILE *out = find_open_file_argument("--out", stdout, "wb", argc, argv);
+    FILE *circuit_file = find_open_file_argument("--circuit", nullptr, "rb", argc, argv);
+    auto circuit = Circuit::from_file(circuit_file);
+    fclose(circuit_file);
+    CircuitStats circuit_stats = circuit.compute_stats();
+
+    const char *types = require_find_argument("--types", argc, argv);
+    bool include_measurements = false, include_detectors = false, include_observables = false;
+    include_measurements = strchr(types, 'M') != nullptr;
+    include_detectors = strchr(types, 'D') != nullptr;
```
Strilanc, line 53: Verify only recognized values are provided.

src/stim/cmd/command_convert.cc
```diff
@@ -0,0 +1,261 @@
...
+        fclose(out);
+    }
+    return EXIT_SUCCESS;
+}
+
+SubCommandHelp stim::command_convert_help() {
+    SubCommandHelp result;
+    result.subcommand_name = "convert";
+    result.description = clean_doc_string(R"PARAGRAPH(
+        Convert data between result formats.
```
Strilanc, line 100: Link the data formats doc on github.

src/stim/cmd/command_convert.test.cc
```diff
@@ -0,0 +1,88 @@
...
+#include "stim/main_namespaced.test.h"
+#include "stim/test_util.test.h"
+
+using namespace stim;
+
+class ConvertTest
+    : public testing::TestWithParam<
+          std::tuple<std::string, std::tuple<std::string, std::string>, std::tuple<std::string, std::string>>> {
+   protected:
+    void SetUp() override {
```
Strilanc, line 26: Avoid coupling data between unit tests like this. Split this up so it's specified separately from each test, even if it's exactly identical in each one. Unit tests are not the place for avoiding repetition (...within reason), they're the place for being straightforward and concrete.

src/stim/cmd/command_convert.test.cc
```diff
@@ -0,0 +1,88 @@
...
+
+std::vector<std::tuple<std::string, std::string>> detection_parameters{
+    std::make_tuple("01", "10000\n11001\n00000\n01001\n"),
+    std::make_tuple("b8", std::string({0x01, 0x13, 0x00, 0x12})),
+    std::make_tuple("dets", "shot D0\nshot D0 D1 L2\nshot\nshot D1 L2\n"),
+    std::make_tuple("hits", "0\n0,1,4\n\n1,4\n"),
+    std::make_tuple("r8", std::string({0x00, 0x04, 0x00, 0x00, 0x02, 0x00, 0x05, 0x01, 0x02, 0x00}))};
+
+TEST_P(ConvertTest, convert) {
+    std::string types = std::get<0>(GetParam());
```
Strilanc, line 54: These tests are simple enough that it'd be clearer to just copy paste the setup rather than trying to share it like this.

src/stim/cmd/command_convert.test.cc
```diff
@@ -0,0 +1,88 @@
...
+#include "stim/test_util.test.h"
+
+using namespace stim;
+
+class ConvertTest
+    : public testing::TestWithParam<
+          std::tuple<std::string, std::tuple<std::string, std::string>, std::tuple<std::string, std::string>>> {
+   protected:
+    void SetUp() override {
+        tmp.write_contents(R"CIRCUIT(
```
Strilanc, line 27: Don't test just one circuit. Use different circuits. Don't share circuits across all the tests like this.

src/stim/main_namespaced.test.cc
```diff
@@ -83,7 +83,7 @@ TEST(main, help_modes) {
         matches(run_captured_stim_main({"--sample", "--repl", "--detect"}, ""), ".+stderr.+More than one mode.+"));
     ASSERT_TRUE(matches(run_captured_stim_main({"--help", "dhnsahddjoidsa"}, ""), ".*Unrecognized.*"));
     ASSERT_TRUE(matches(run_captured_stim_main({"--help", "H"}, ""), ".+Hadamard.+"));
-    ASSERT_TRUE(matches(run_captured_stim_main({"--help", "--sample"}, ""), ".*Samples measurements from a circuit.+"));
+    ASSERT_TRUE(matches(run_captured_stim_main({"--help", "--sample"}, ""), ".*Unrecognized.*"));
```
Strilanc, line 86: Why did this change?

src/stim/cmd/command_convert.cc
justinledford: done

src/stim/cmd/command_convert.cc
justinledford: done

src/stim/cmd/command_convert.test.cc
justinledford: done

src/stim/cmd/command_convert.test.cc
justinledford: done

justinledford:
> 



> This is a good start, but there are several more use cases that I think need to be covered. Extend this to include the following use cases:
> 
> * I have the DEM, not the circuit, and want to convert detection event data.
> * I have neither the DEM nor the circuit but I know all the values that would be derived from them like the bits per shot
> * I know how many bits per shot, but I don't know if it's detection event data or measurement data (and I'm not using the DETS format which is the only one where that distinction matters)
> * The input file has observable data appended to the detector data, and I want to split the observable data out into a separate file
> 
> These use cases all kind of differ in what information they need. In general the command should infer everything it can from whatever the user provides, and if it can't infer a needed quantity it should raise an error identifying that quantity and saying various ways it can be provided.

Thanks, that is a very good point. I naively assumed the user would always have the circuit on hand. I added support for all of these cases, but there is another I'm wondering if we should add. For DETS files we can also attempt to parse the text looking for the value prefixes, but this may not be reliable as the file would need to include the max index of each type, which may not always occur. What do you think?

src/stim/main_namespaced.test.cc
Strilanc: That's probably because nullptr is being used to indicate no input, so the corresponding type would be optional<string> not string. My recommendation is to leave it as a const pointer. Seeing a test change failure symptoms like this is a huge red flag on a refactor of a method.

Strilanc:
Because it's not reliable to infer sizes or types from a det file, I would not attempt to do so.

src/stim/main_namespaced.test.cc
justinledford: Right that's why I made sure to understand to cause of the failure. The confusing thing is that the current output of `--help --sample` does not match the test case:

```
$ ./bazel-bin/stim --help --sample
Unrecognized help topic '--sample'.
```

So even if we keep the function as taking a const pointer, shouldn't we update the test case to properly reflect the real output of that command?

justinledford: (note the above output is on a freshly cloned instance of the repo built at HEAD)

Strilanc: Hmmmmm... that does seem wrong. It looks like `out/stim --help --sample --in UNUSED` is giving output that matches the string but it's part of listing all commands instead of specifically about sampling.

src/stim/cmd/command_convert.cc
```diff
@@ -0,0 +1,567 @@
...
+}
+
+void process_dem(const char *dem_path, DataDetails *details_out) {
+    if (dem_path == nullptr) {
+        return;
+    }
+
+    FILE *dem_file = fopen(dem_path, "rb");
+    if (dem_file == nullptr) {
+        std::stringstream msg;
```
Strilanc, line 56: You'd get this checking for free from `find_open_file_argument("--dem", nullptr, "rb", argc, argv)`

src/stim/cmd/command_convert.cc
```diff
@@ -0,0 +1,567 @@
...
+void process_circuit(const char *circuit_path, const char *types, DataDetails *details_out) {
+    if (circuit_path == nullptr) {
+        return;
+    }
+    if (types == nullptr) {
+        throw std::invalid_argument("--types required when passing circuit");
+    }
+    FILE *circuit_file = fopen(circuit_path, "rb");
+    if (circuit_file == nullptr) {
+        std::stringstream msg;
```
Strilanc, line 77: `find_open_file_argument`

src/stim/cmd/command_help.cc
```diff
@@ -90,6 +91,7 @@ std::vector<SubCommandHelp> make_sub_command_help() {
     help_help.description = "Prints helpful information about using stim.";
     auto result = std::vector<SubCommandHelp>{
         command_analyze_errors_help(),
+        command_convert_help(),
```
Strilanc, line 94: You need to run the regen docs command in the dev/ directory to include this in the docs in the repo. (It assumes you're in a python virtual env with the dev wheel installed; see the dev docs.)

Strilanc (changes requested):
Look pretty good, just a few things left I think

src/stim/cmd/command_convert.cc
justinledford: I actually tried this originally, but `find_open_file_argument` throws an exception if this arg isn't passed (and this arg is optional): https://github.com/quantumlib/Stim/blob/60baacac26dfcc2a1e628f7426cc290d3d252266/src/stim/arg_parse.cc#L336-L340

We could extend `find_open_file_argument` to deal with optional FILE args if you think that'd be useful.

Strilanc: Yeah that'd be pretty useful. Not needed for this PR, but in general the coupling of the default value to does-it-throw has been a thorn in those methods' utility.

justinledford: sure, I can follow up in a subsequent PR to add this and clean up any other uses where we have optional file args

src/stim/cmd/command_help.cc
justinledford: done

PR #603, opened 2023-08-10
Add left/right shift and addition to simd_bits.

fdmalone:
From #598 added +=, >>= and <<= to simd_bits. It wasn't obvious to me that these could use word level parallelism without using more memory? For example, the shifts could store the relevant carry masks and or these at the end but this would require a temporary of the same size as the simd_bits instance.


Strilanc (changes requested):

Strilanc:
Looks good but let's tighten the tests with some fuzz testing and some corner cases like shifting by 0.

src/stim/mem/simd_bits_range_ref.test.cc
```diff
@@ -162,6 +162,77 @@ TEST_EACH_WORD_SIZE_W(simd_bits_range_ref, equality, {
     ASSERT_FALSE(m0 != m1);
 })
 
+TEST_EACH_WORD_SIZE_W(simd_bits_range_ref, add_assignment, {
+    alignas(64) std::array<uint64_t, 8> data{
```
Strilanc, line 166: oooo I kinda like this way of setting up the test data.

src/stim/mem/simd_bits.test.cc
```diff
@@ -253,6 +317,23 @@ TEST_EACH_WORD_SIZE_W(simd_bits, left_shift_assignment, {
     }
     m0 <<= 488;
     ASSERT_TRUE(!m0.not_zero());
+    simd_bits<W> m1(448), m2(448);
```
Strilanc, line 320: Separate the fuzz test out into a test with "fuzz" in the name, and put the body in a loop that repeats 5 times (increase it to 1000 while you write it but for detecting regressions 5 is more than enough IMO).

src/stim/mem/simd_bits.test.cc
```diff
@@ -253,6 +317,23 @@ TEST_EACH_WORD_SIZE_W(simd_bits, left_shift_assignment, {
...
+    m1.randomize(448, SHARED_TEST_RNG());
+    m2 = m1;
+    m1 <<= 1;
+    for (size_t k = 0; k < 448 - 1; k++) {
+        ASSERT_EQ(m1[k + 1], m2[k]);
+    }
+    m1.randomize(448, SHARED_TEST_RNG());
+    m2 = m1;
+    m1 <<= 10;
```
Strilanc, line 329: Also try a random offset from [0, num_bits_padded()] and verify bits up to the padding not just up to the requested length.

src/stim/mem/simd_bits.test.cc
Strilanc, line 320: Also randomize requested length from 1 to 1200.

Strilanc (changes requested):

src/stim/mem/simd_bits.test.cc
```diff
@@ -274,32 +281,36 @@ TEST_EACH_WORD_SIZE_W(simd_bits, right_shift_assignment, {
...
+        std::uniform_int_distribution dist_bits(1, 1200);
+        int num_bits = dist_bits(rng);
+        simd_bits<W> m1(num_bits), m2(num_bits);
+        m1.randomize(num_bits, rng);
+        m2 = m1;
+        std::uniform_int_distribution dist_shift(0, (int)m1.num_bits_padded());
+        size_t shift = dist_shift(rng);
+        m1 >>= shift;
+        for (size_t k = 0; k < m1.num_bits_padded() - shift; k++) {
+            ASSERT_EQ(m1[k], m2[k + shift]);
```
Strilanc, line 298: I'd also check the shifted in bits. That said I basically trust the method at this point. Was just worried about the corner cases that come up.

Strilanc (approved):

src/stim/mem/simd_bits.test.cc
fdmalone: Do you mean in m1 check that the bits from num_bits_padded() - shift to num_bits_padded() are zero?

Strilanc: Yeah check every single output bit.

PR #618, opened 2023-08-22
Increase flexibility of stim.Circuit.diagram filter_coords args

Strilanc:
- Fix the fact that it doesn't take DemTarget instances
- Allow passing strings like "D5" and "L0"
- Allow passing an individual filter instead of a list
- Allow saying 'detslice' instead of 'detslice-svg'

Fixes https://github.com/quantumlib/Stim/issues/616


Strilanc:
@mmcewen-g

Issue #623, opened 2023-08-24
Can pybind11 version be updated / relaxed?


tjni:
Hello! I am working on packaging in [nixpkgs](https://github.com/NixOS/nixpkgs), and we have updated to pybind11 2.11.1. I am sorry I know nothing about pybind11. Is it a dependency whose version can be relaxed to support this?

Strilanc:
Are you referring to the version declared in `WORKSPACE` (for the Bazel build) or in `pyproject.toml` (for the setup.py build)?

The Bazel one is quite hard to loosen because e.g. it declares a sha256 hash. The setup.py one can probably be changed from `==2.9.2` to `~=2.11`.

tjni:
Thank you for the fast reply! I'm referring to the `pyproject.toml` one. If you don't mind, I can create a PR and we can see.

Strilanc:
Go for it.

PR #624, opened 2023-08-24
Update pybind11 to ~=2.11

tjni:
Tries to update `pybind11` to the latest version to see what happens.

Could close https://github.com/quantumlib/Stim/issues/623.


tjni:
I'm so sorry, I didn't realize this was a Google product and requires signing a CLA. For personal reasons, I don't really want to do that. I don't think change is really copyrightable, but I understand there's process. I'm going to close this PR due to that.

PR #633, opened 2023-09-05
Fix bug with simd_bits +=.

fdmalone:
Caught this when trying to address #598.

### DIFF
--- a/src/stim/mem/simd_bits.test.cc
+++ b/src/stim/mem/simd_bits.test.cc
@@ -164,6 +164,15 @@ TEST_EACH_WORD_SIZE_W(simd_bits, xor_assignment, {
     }
 })
 
+template <size_t W>
+void set_bits_from_u64_vector(simd_bits<W> &bits, std::vector<uint64_t> &vec) {
+    for (size_t w = 0; w < bits.num_u64_padded(); w++) {
+        for (size_t b = 0; b < 64; b++) {
+            bits[w * 64 + b] |= (vec[w] & (1ULL << b));
+        }
+    }
+}
+
 TEST_EACH_WORD_SIZE_W(simd_bits, add_assignment, {
     simd_bits<W> m0(512);
     simd_bits<W> m1(512);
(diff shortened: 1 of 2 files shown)


fdmalone:
Windows test is failing to build cirq?

src/stim/mem/simd_bits.test.cc
```diff
@@ -228,6 +228,22 @@ TEST_EACH_WORD_SIZE_W(simd_bits, add_assignment, {
...
+        for (size_t k = 0; k < 64; k++) {
+            add[word * 64 + k] = 1;
+        }
+    }
+    one[0] = 1;
+    add += one;
+    // These should all overflow and carries should propagate to the last word.
+    for (size_t k = 0; k < num_bits - 1; k++) {
+        ASSERT_EQ(add[k], 0);
+    }
```
Strilanc, line 245: Given that this failed, please add a fuzz test. Bias the fuzz test towards interesting cases by with uniform 1/3 chance for each word making it all-1 or all-0 or all-random.

Strilanc (changes requested):

Strilanc:
Looks like a hash failed to match. Probably transient, and if not then it's matplotlib's problem not yours.

src/stim/mem/simd_bits.test.cc
fdmalone: Is testing by adding bit-by-bit using [] sufficient here, or is testing against python preferred, second way requires exposing to pybind?

Strilanc: Ah, fair point, you need a ground truth to compare against, and it has to come from code that's unlikely to have the same bugs.... hmmm....

Instead of checking the exact answer, you could check properties like commutativity and associativity. You could verify that `a += x; a = ~a; a += x; a = ~a` results in `a` being unchanged. You could check that `a + a == a << 1` and that `a + ~a == 1`. You could check that `a + (b & c) + (b ^ c) == a + b + c`. Anything that sortof rearranges the problem in a way that would trigger the boundary issues differently on the left hand side and right hand side.

And yeah, pulling in a few hardcoded cases from python would also work. I wouldn't expose simd_bits all the way up to python.

fdmalone: Cool, good suggestions!

fdmalone:
For the fuzz tests I ran 10000 times initially. I also checked with the following python code:

``` python
from typing import List, Union
import numpy as np
from fixedint import UInt64

def join_words(words: List[int]) -> int:
    """Join a list of 64bit words into a single python integer"""
    out_str = ''
    for x in words[::-1]:
        out_str += f"{x:064b}"
    return int(out_str, 2)

def split_into_words(x: str) -> List[int]:
    """Split a binary representation of a python int into a list of 64 bit words"""
    strings = [x[i:i+64] for i in range(0, len(x), 64)]
    return [int(i, 2) for i in strings[::-1]]

def adder(lhs: List[UInt64], rhs: List[UInt64]) -> List[UInt64]:
    """Multi-word adder"""
    carry = 0
    for i, _ in enumerate(lhs):
        vbf = lhs[i]
        lhs[i] = lhs[i] + rhs[i] + carry
        carry = lhs[i] < vbf or (carry & (lhs[i] == vbf))
    return lhs


def to_fixed(x: int, width=192):
    """Fixed width string rep of python integer"""
    return f"{x:0{width}b}"[-width:]

def to_cpp_vec(x: List[Union[int, UInt64]], label="x"):
    """Write as a vector of integers"""
    out_str = f"std::vector<uint64_t> {label}{{"
    for i, w in enumerate(x):
        comma = "," if i < len(x) - 1 else ""
        out_str += f"{w}ULL{comma}"
    out_str += "};"
    print(out_str)

def gen_random_integer(num_bits):
    num_words = (num_bits - 1) // 64 + 1
    words = []
    for iw in range(num_words):
        r = np.random.random()
        b = 0
        if r < 1/3:
            b = 0
        elif r > 2/3:
            if iw == num_words - 1:
                for k in range(num_bits - 64 * iw):
                    b |= (1<< k)
            else:
                b = 0xFFFFFFFFFFFFFFFF
        else:
            for k in range(64):
                if np.random.random() > 0.5:
                    b |= (1<< k)
        words.append(b)
    return words

if __name__ == "__main__":
    np.random.seed(7)
    #for i in range(10_000):
    for i in range(4):
        #num_bits = np.random.randint(10, 1200) 
        num_bits = 512
        x = gen_random_integer(num_bits)
        y = gen_random_integer(num_bits)
        a = join_words(x)
        b = join_words(y)
        xpy = to_fixed(a+b, num_bits)
        xu = [UInt64(i) for i in x]
        yu = [UInt64(i) for i in y]
        z = adder(xu, yu)
        z_str = to_fixed(join_words(z), num_bits)
        xpy_str = xpy
        if i == 3:
            to_cpp_vec(x, "x")
            to_cpp_vec(y, "y")
            to_cpp_vec(split_into_words(xpy), "z")
        assert to_fixed(join_words(z), num_bits) == xpy
        assert xpy_str == z_str
```

Strilanc:
The tests appear to be crashing with buffer overflows.

fdmalone:
The perils of developing on my mac

src/stim/mem/simd_bits.test.cc
```diff
@@ -228,6 +237,149 @@ TEST_EACH_WORD_SIZE_W(simd_bits, add_assignment, {
...
+        simd_bits<W> m1(num_bits);
+        simd_bits<W> allset(num_bits);
+        allset.invert_bits();
+        set_random_words_to_all_set(m1, num_bits, rng, dist_real);
+        simd_bits<W> m2(m1);
+        m2.invert_bits();
+        m1 += m2;
+        ASSERT_EQ(m1, allset);
+    }
+    // m1 += x; m1 = ~x; m1 += x; m1 is unchanged.
```
Strilanc, line 346: m1 = ~m1 not m1 = ~x

src/stim/mem/simd_bits.test.cc
```diff
@@ -228,6 +237,149 @@ TEST_EACH_WORD_SIZE_W(simd_bits, add_assignment, {
...
+        int num_bits = dist_bits(rng);
+        simd_bits<W> m1(num_bits), m2(num_bits);
+        set_random_words_to_all_set(m1, num_bits, rng, dist_real);
+        set_random_words_to_all_set(m2, num_bits, rng, dist_real);
+        simd_bits<W> ref1(m1), ref2(m2);
+        m1 += ref2;
+        m2 += ref1;
+        ASSERT_EQ(m1, m2);
+    }
+    // a + ~a = allset
```
Strilanc, line 333: Change this one to `a + ~a + 1 == 0`. That way you for sure trigger all the carries.

Strilanc (changes requested):
I think this is the last round of review; feeling much more confident about it now.

Strilanc (approved):

PR #667, opened 2023-11-22
Update python_api_reference_vDev.md

tomirendo:
Fixed a typo

### DIFF
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -1309,7 +1309,7 @@ def detector_error_model(
             error mechanisms). When set to true, the probabilities of the disjoint
             cases are instead assumed to be independent probabilities. For example,
             a `PAULI_CHANNEL_1(0.1, 0.2, 0.0)` becomes equivalent to an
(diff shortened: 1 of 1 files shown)


Strilanc:
I agree this is a typo. You've edited it in a generated file, instead of in the source that generates it. If you search for

```
`X_ERROR(0.1)` followed by a `Z_ERROR(0.2)`. This assu
```

across the entire code base, you'll find the other places it has to change. Technically it only has to change in the one place under the `src/` directory, at which point `dev/regen_docs.sh` will do all of the others.

(Potentially also interesting to this particular doc string is that the last version of stim introduced code that attempts to do perfect conversions of `PAULI_CHANNEL_1` when possible. The given example is still correct because there's no perfect conversion when one of the three probabilities is set to 0. I think it's fine as is.)

tomirendo:
Hey Craig, thanks for responding so quickly! I'll rerun the script and push again. 
I should learn to do these things by now. 
(Also exciting about the decomposition code, I'll comment that as well)

Strilanc:
Merged into https://github.com/quantumlib/Stim/pull/721

Issue #653, opened 2023-11-14
add tools for manipulating `stim.FlipSimulator` in parallel


ghost:
`stim.FlipSimulator.set_pauli_flips(paulis: Iterable[int | str], qubit: int)` w `len(paulis) == batch_size`

`stim.FlipSimulator.xor_pauli_flips(paulis: Iterable[int | str], qubit: int)` w `len(paulis) == batch_size`
to avoid needing to peek and set

`stimFlipSimulator.apply_pauli_error(pauli: str|int, mask: Iterable[bool])` w `len(mask) == batch_size * qubits`
for the more common case of wanting to apply 1 pauli error, but selectively over both qubits and simulation indices
(might be more clever ways of providing the ~2D mask other than just flattening it, could do `Iterable[Iterable[bool]]`, but then it's harder to prevent ragged data / take advantage of it being rectangular)

Strilanc:
Done now

PR #678, opened 2023-12-07
Add `stim.FlipSimulator.broadcast_pauli_errors`

mmcewen-g:
Addresses part of #653

Gives a general method for applying errors with 2D masks, key for speeding up interactive use of the flip simulator

### DIFF
--- a/src/stim/simulators/frame_simulator.pybind.cc
+++ b/src/stim/simulators/frame_simulator.pybind.cc
@@ -786,17 +786,18 @@ void stim_pybind::pybind_frame_simulator_methods(
                     if (s == "X") {
                         p = 1;
                     } else if (s == "Y") {
-                        p = 3;
-                    } else if (s == "Z") {
                         p = 2;
+                    } else if (s == "Z") {
+                        p = 3;
                     } else if (s == "I" || s == "_") {
                         p = 0;
                     }
                 } catch (const pybind11::cast_error &) {
                 }
             }
-            bool flip_x_part = p & 1;
+
             bool flip_z_part = p & 2;
+            bool flip_x_part = 6 >> p & 1;  // 0b0110 >> p & 0b0001
 
             if (pybind11::isinstance<pybind11::array_t<bool>>(mask)) {
                 // can use pybind11::isinstance<pybind11::array_t<bool, pybind11::array::c_style>>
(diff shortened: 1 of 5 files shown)


src/stim/simulators/frame_simulator.pybind.cc
```diff
@@ -770,4 +770,91 @@ void stim_pybind::pybind_frame_simulator_methods(
...
+                            self.x_table[i][j] ^= *b & flip_x_part;
+                            self.z_table[i][j] ^= *b & flip_z_part;
+                        }
+                    }
+                } else {
+                    throw std::invalid_argument(
+                    "apply_pauli_errors currently only supports a mask that is a 2D bool numpy array");
+                }
+            } else {
+                    throw std::invalid_argument(
```
Strilanc, line 825: Try to merge the two throws since they're the same error message.

src/stim/simulators/frame_simulator_pybind_test.py
```diff
@@ -216,6 +216,98 @@ def test_set_pauli_flip():
         stim.PauliString('XZ___'),
     ]
 
+def test_apply_pauli_errors():
+    sim = stim.FlipSimulator(
```
Strilanc, line 220: Check corner and invalid cases. Try `pauli='I'` and `pauli=0` and `pauli="whoops"`

src/stim/simulators/frame_simulator.pybind.cc
mmcewen-g: Not sure how to do this, given i need to do the dim=2 check after i've done the cast, which means ive already done the is_array check; I could change the reference to a pointer so i can have it be NULL before I do the cast? (not sure if this is frowned upon in c++ land) or i could do the cast check twice, or i could allocate extra variables,
Seems like more of a style thing here, so happy to be told what to do

Strilanc: Not a big deal if it's hard to do.

src/stim/simulators/frame_simulator.pybind.cc
```diff
@@ -770,4 +770,102 @@ void stim_pybind::pybind_frame_simulator_methods(
...
+            const pybind11::array_t<bool> &arr = pybind11::cast<pybind11::array_t<bool>>(mask);
+
+            if (arr.ndim() != 2) {
+                throw std::invalid_argument(
+                "broadcast_pauli_errors can only accept mask that is a 2D array of np.bool_");
+            }
+
+            size_t major = arr.shape(0);
+            size_t minor = arr.shape(1);
+            auto u = arr.unchecked<2>();
```
Strilanc, line 823: You didn't check that the size of the array is correct.

For the qubit axis, you should be lenient. This is consistent with the simulator auto-expanding when new qubits are mentioned. Call `self.ensure_safe_to_do_circuit_with_stats(CircuitStats{.num_qubits=X})` to ensure the simulation state is not too small. Continue to iterate up to the array's qubit count, which may be less than the simulator's qubit count.

For the instance count, you should be strict. This is where the speed comes from. If the sim's batch size differs from the array shape, throw an exception.

src/stim/simulators/frame_simulator_pybind_test.py
```diff
@@ -216,6 +216,143 @@ def test_set_pauli_flip():
...
+            [True, False],
+            [False, False]]
+        ),
+    )
+    peek = sim.peek_pauli_flips()
+    assert peek == [
+        stim.PauliString("+YZX"),
+        stim.PauliString("+Z_X")
+    ]
+    sim.broadcast_pauli_errors(
```
Strilanc, line 251: Add tests with arrays that have fewer and more qubit indices than the current simulator state.

Add tests that array with wrong batch size raise an exception.

src/stim/simulators/frame_simulator_pybind_test.py
```diff
@@ -216,6 +216,143 @@ def test_set_pauli_flip():
...
+            [True, True]]
+        ),
+    )
+    peek = sim.peek_pauli_flips()
+    assert peek == [
+        stim.PauliString("+_ZX"),
+        stim.PauliString("+ZYZ")
+    ]
+
+    with pytest.raises(Exception):
```
Strilanc, line 337: Specify `ValueError` instead of just `Exception` and add a `match="pauli"` to confirm it's at least vaguely the expected error.

src/stim/simulators/frame_simulator_pybind_test.py
```diff
@@ -216,6 +216,143 @@ def test_set_pauli_flip():
...
+        sim.broadcast_pauli_errors(
+            pauli='whoops',
+            mask=np.asarray([
+                [True, True],
+                [False, True],
+                [True, True]]
+            ),
+        )
+    with pytest.raises(Exception):
```
Strilanc, line 346: Specify `ValueError` instead of just `Exception` and add a `match="pauli"` to confirm it's at least vaguely the expected error.

Strilanc (changes requested):
Gettin closer

src/stim/simulators/frame_simulator.pybind.cc
```diff
@@ -820,6 +820,14 @@ void stim_pybind::pybind_frame_simulator_methods(
 
             size_t major = arr.shape(0);
             size_t minor = arr.shape(1);
+
+            if (minor != self.batch_size) {
+                throw std::invalid_argument(
+                "broadcast_pauli_errors can only accept mask that has minor shape equal to the batch_size");
+            }
+
+            self.ensure_safe_to_do_circuit_with_stats(CircuitStats{.num_qubits=(uint32_t)major});
```
Strilanc, line 829: This cast isn't safe. What if the shape is larger than 2^32? Do a bounds check first. Gotta make sure we never segfault people's python instances in their notebooks; they hate that.

Strilanc (changes requested):
One last safety check and we're good to go.

Strilanc (approved):

PR #692, opened 2024-02-05
Add `CPP` gate

Strilanc:


doc/gates.md
```diff
@@ -2976,6 +2977,72 @@ Decomposition (into H, S, CX, M, R):
...
+
+    # Swap qubit 1 and qubit 5 by negating their overlap with the singlet state.
+    CPP X1*X5 Z1*Z5
+
+    # Negate the amplitude of the |00> state.
+    CPP !Z0 !Z1
+
+    # Measure qubit 0 and do Pauli operations conditioned on the measurement returning TRUE.
+    M 0
+    CPP rec[-1] X1*Y2*Z3
```
perlinm, line 3019: Is something like `CPP rec[-1]*rec[-2] X1*X2*X3` allowed?

Here I would interpret `rec[-1]*rec[-2]` as the parity of two measurements.

Strilanc: Yes, that would work and that would be the semantics.

PR #703, opened 2024-02-28
Add `stim.Circuit.{shortest,likeliest}_error_sat_problem`

noajshu:
## Summary
Adds a generator that outputs a .wcnf file string in [WDIMACS format](http://www.maxhs.org/docs/wdimacs.html).
Added a few tests as well.

## Demo for the surface code
Here is a demo:

**Step 1:** Generate the WCNF file
```python
import stim
c = stim.Circuit.generated(
            "surface_code:rotated_memory_z",
            distance=5,
            rounds=2,
            after_clifford_depolarization=0.001,
)
wcnf = c.shortest_error_sat_problem()
with open('d5problem.wcnf', 'w') as outfile:
    outfile.write(wcnf)
```

**Step 2:** Fetch a solver from the [2023 maxSAT competition](https://maxsat-evaluations.github.io/2023/)
```bash
wget https://maxsat-evaluations.github.io/2023/mse23-solver-src/exact/CASHWMaxSAT-CorePlus.zip
unzip CASHWMaxSAT-CorePlus.zip
```

**Step 3:** Run the solver on the produced file
```bash
time ./CASHWMaxSAT-CorePlus/bin/cashwmaxsatcoreplus -bm -m d5problem.wcnf 
c Parsing MaxSAT file...
c Using SCIP solver, version 8.0.3, https://www.scipopt.org
c scip_time = 0.000000
c Starting SCIP solver in a separate thread (with time-limit = 0s) ...
c Using COMiniSatPS SAT solver by Chanseok Oh (2016)
o 5
c ============================[  Problem Statistics ]============================
c |  Number of variables:           995                                         |
c |  Number of clauses:            4377 (incl.            5 soft in queue)      |
c ===============================================================================
c Optimal solution: 5
c _______________________________________________________________________________
c 
c restarts               : 881
c conflicts              : 88351          (20350 /sec)
c decisions              : 257865         (0.00 % random) (59394 /sec)
c propagations           : 6079046        (1400176 /sec)
c conflict literals      : 4980307        (6.84 % deleted)
c =======================[ UWrMaxSat resources usage ]===========================
c Memory used            : 164.00 MB
c CPU time               : 4.34163 s
c Wall clock time        : 3 s
c OptExp Enc: Srt/BDD/Add: 12 0 0
c _______________________________________________________________________________
v 0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000110010100001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010100000100010000010010000000000001
s OPTIMUM FOUND

real	0m2.215s
user	0m4.345s
sys	0m0.051s
```

## Verifying distances of several quasicyclic codes
I ran this on some circuits from [this paper](https://arxiv.org/pdf/2308.07915.pdf), with help from @oscarhiggott to generate the circuits.
The log file is attached below:
[n72_n90_n103_quasicyclic_codes_have_expected_distances_6_8_8.txt](https://github.com/quantumlib/Stim/files/14427119/n72_n90_n103_quasicyclic_codes_have_expected_distances_6_8_8.txt)
So far, it has proven that the n=72, n=29, and n=103 circuits have the expected distance reported in [the paper](https://arxiv.org/pdf/2308.07915.pdf).
I am still running solvers on the n=144 and n=288 cases. My computer got rebooted so I had to start over. I am cautiously optimistic that the solvers will finish within a few weeks.

### DIFF
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -2447,7 +2447,7 @@ def shortest_error_problem_as_wcnf_file(
     self,
     *,
     weighted: bool = False,
-    num_distinct_weights: int = 1,
+    weight_scale_factor: int = 0,
 ) -> str:
     """Generates a maxSAT problem instance in WDIMACS format whose optimal value is
     the distance of the protocol, i.e. the minimum weight of any set of errors
@@ -2466,16 +2466,17 @@ def shortest_error_problem_as_wcnf_file(
         weighted: Defaults to False (unweighted), so that the problem is to find
         the minimum size set of errors that leads to an undetectable logical
         error.
-        num_distinct_weights: Defaults to 1. If weighted==False, must set
-        num_distinct_weights to 1. If weighted==True, num_distinct_weights can be
-        set to a value > 1. For a weighted problem, the errors will be quantized
-        into at most this many distinct weight values and the sum of the weights
+        weight_scale_factor: Defaults to 0. If weighted==False, must set
+        weight_scale_factor to 0. If weighted==True, weight_scale_factor must be
+        set to a value >= 1. For a weighted problem, the errors will be quantized
+        into integers using this scale factor and the sum of the weights
         will be minimized rather than the cardinality of the set of errors. For
-        a reasonably large quantization (num_distinct_weights > 100), the .wcnf
-        file solution should be the (approximately) most likely undetectable
-        logical error. Note, however, that maxSAT solvers often become slower
-        when many distinct weights are provided, so for computing the distance
-        it is better to use the default quantization num_distinct_weights = 1.
+        a reasonably large quantization (weight_scale_factor > 100 is often
+        sufficient), the .wcnf file solution should be the (approximately)
+        most likely undetectable logical error. Note, however, that maxSAT solvers
+        can become slower when many distinct weights are provided, so for computing
+        the distance it is better to simply use the defaults weighted=False and
+        weight_scale_factor = 0.
 
     Returns:
         A WCNF file in [WDIMACS format](http://www.maxhs.org/docs/wdimacs.html)
@@ -2483,12 +2484,12 @@ def shortest_error_problem_as_wcnf_file(
     Examples:
         >>> import stim
         >>> circuit = stim.Circuit("""
-        ...     X_ERROR(0.1) 0
-        ...     M 0
-        ...     OBSERVABLE_INCLUDE(0) rec[-1]
-        ...     X_ERROR(0.4) 0
-        ...     M 0
-        ...     DETECTOR rec[-1] rec[-2]
+        ...   X_ERROR(0.1) 0
+        ...   M 0
+        ...   OBSERVABLE_INCLUDE(0) rec[-1]
+        ...   X_ERROR(0.4) 0
+        ...   M 0
+        ...   DETECTOR rec[-1] rec[-2]
         ... """)
         >>> print(circuit.shortest_error_problem_as_wcnf_file())
         p wcnf 2 4 5
@@ -2497,16 +2498,15 @@ def shortest_error_problem_as_wcnf_file(
         5 -1 0
         5 2 0
 
-        >>> 
         >>> print(circuit.shortest_error_problem_as_wcnf_file(
         ...   weighted=True,
-        ...   num_distinct_weights=10
+        ...   weight_scale_factor=1000
         ... ))
-        p wcnf 2 4 41
-        2 -1 0
-        10 -2 0
-        41 -1 0
-        41 2 0
+        p wcnf 2 4 4001
+        185 -1 0
+        1000 -2 0
+        4001 -1 0
+        4001 2 0
     """
 ```
 
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -2145,13 +2145,13 @@ def likeliest_error_sat_problem(
     quantization: int = 100,
     format: str = 'WDIMACS',
 ) -> str:
-    """Makes a maxSAT problem of the circuit's distance, that other tools can solve.
+    """Makes a maxSAT problem of the circuit's most likely undetectable logical
+    error, that other tools can solve.
 
     The output is a string describing the maxSAT problem in WDIMACS format
     (see https://maxhs.org/docs/wdimacs.html). The optimal solution to the
-    problem is the fault distance of the circuit (the minimum number of error
-    mechanisms that combine to flip any logical observable while producing no
-    detection events).
+    problem is the highest likelihood set of error mechanisms that combine to
+    flip any logical observable while producing no detection events).
 
     There are many tools that can solve maxSAT problems in WDIMACS format.
     For example, you can download one of the entries in the 2023 maxSAT
@@ -2165,12 +2165,12 @@ def likeliest_error_sat_problem(
     Args:
         format: Defaults to "WDIMACS", corresponding to WDIMACS format which is
         described here: http://www.maxhs.org/docs/wdimacs.html
-        weight_bins: Defaults to 1 (unweighted errors). Error probabilities are
-            converted to log-odds and scaled/rounded to be positive integers at
-            most this large. Setting this argument to a larger number results in
-            the solution to the problem being the most likely logical error
-            weighted by probability of its parts, instead of just the one with
-            the fewest parts.
+        quantization: Defaults to 10. Error probabilities are converted to log-odds
+            and scaled/rounded to be positive integers at most this large. Setting
+            this argument to a larger number results in more accurate quantization
+            such that the returned error set should have a likelihood closer to the
+            true most likely solution. This comes at the cost of making some maxSAT
+            solvers slower.
 
     Returns:
         A string corresponding to the contents of a WCNF file in the requested
@@ -2541,12 +2541,6 @@ def shortest_error_sat_problem(
     Args:
         format: Defaults to "WDIMACS", corresponding to WDIMACS format which is
         described here: http://www.maxhs.org/docs/wdimacs.html
-        weight_bins: Defaults to 1 (unweighted errors). Error probabilities are
-            converted to log-odds and scaled/rounded to be positive integers at
-            most this large. Setting this argument to a larger number results in
-            the solution to the problem being the most likely logical error
-            weighted by probability of its parts, instead of just the one with
-            the fewest parts.
 
     Returns:
         A string corresponding to the contents of a WCNF file in the requested
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -2153,6 +2153,12 @@ def likeliest_error_sat_problem(
     problem is the highest likelihood set of error mechanisms that combine to
     flip any logical observable while producing no detection events).
 
+    If there are any errors with probability p > 0.5, they are inverted so
+    that the resulting weight ends up being positive. If there are errors
+    with weight close or equal to 0.5, they can end up with 0 weight meaning
+    that they can be included or not in the solution with no affect on the
+    likelihood.
+
     There are many tools that can solve maxSAT problems in WDIMACS format.
     One quick way to get started is to install pysat by running this BASH
     terminal command:
@@ -2535,7 +2541,9 @@ def shortest_error_sat_problem(
     (see https://maxhs.org/docs/wdimacs.html). The optimal solution to the
     problem is the fault distance of the circuit (the minimum number of error
     mechanisms that combine to flip any logical observable while producing no
-    detection events).
+    detection events). This method ignores the probabilities of the error
+    mechanisms since it only cares about minimizing the number of errors
+    triggered.
 
     There are many tools that can solve maxSAT problems in WDIMACS format.
     One quick way to get started is to install pysat by running this BASH
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -46,6 +46,7 @@ API references for stable versions are kept on the [stim github wiki](https://gi
     - [`stim.Circuit.num_ticks`](#stim.Circuit.num_ticks)
     - [`stim.Circuit.reference_sample`](#stim.Circuit.reference_sample)
     - [`stim.Circuit.search_for_undetectable_logical_errors`](#stim.Circuit.search_for_undetectable_logical_errors)
+    - [`stim.Circuit.shortest_error_problem_as_wcnf_file`](#stim.Circuit.shortest_error_problem_as_wcnf_file)
     - [`stim.Circuit.shortest_graphlike_error`](#stim.Circuit.shortest_graphlike_error)
     - [`stim.Circuit.to_file`](#stim.Circuit.to_file)
     - [`stim.Circuit.to_qasm`](#stim.Circuit.to_qasm)
@@ -2434,6 +2435,55 @@ def search_for_undetectable_logical_errors(
         ...     dont_explore_edges_increasing_symptom_degree=True,
         ... )))
         5
+    """
+```
+
+<a name="stim.Circuit.shortest_error_problem_as_wcnf_file"></a>
+```python
+# stim.Circuit.shortest_error_problem_as_wcnf_file
+
+# (in class stim.Circuit)
+def shortest_error_problem_as_wcnf_file(
+    self,
+    *,
+    weighted: bool = False,
+    num_distinct_weights: int = 1,
+) -> str:
+    """Generates a maxSAT problem instance in WDIMACS format whose optimal value is
+    the distance of the protocol, i.e. the minimum weight of any set of errors
+    that forms an undetectable logical error.
+
+    Args:
+        weighted: Defaults to False (unweighted), so that the problem is to find
+        the minimum size set of errors that leads to an undetectable logical
+        error.
+        num_distinct_weights: Defaults to 1. If weighted==False, must set
+        num_distinct_weights to 1. If weighted==True, num_distinct_weights can be
+        set to a value > 1. For a weighted problem, the errors will be quantized
+        into at most this many distinct weight values and the sum of the weights
+        will be minimized rather than the cardinality of the set of errors. For
+        a reasonably large quantization (num_distinct_weights > 100), the .wcnf
+        file solution should be the (approximately) most likely undetectable
+        logical error. Note, however, that maxSAT solvers often become slower
+        when many distinct weights are provided, so for computing the distance
+        it is better to use the default quantization num_distinct_weights = 1.
+
+    Returns:
+        A WCNF file in [WDIMACS format](http://www.maxhs.org/docs/wdimacs.html)
+
+    Examples:
+        >>> import stim
+        >>> circuit = stim.Circuit.generated(
+        ...     "surface_code:rotated_memory_x",
+        ...     rounds=5,
+        ...     distance=5,
+        ...     after_clifford_depolarization=0.001)
+        >>> print(circuit.shortest_undetectable_logical_error_wcnf(
+                        num_distinct_weights=1))
+        ....
+        >>> print(circuit.shortest_undetectable_logical_error_wcnf(
+                        num_distinct_weights=10))
+        ....
     """
 ```
 
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -2154,17 +2154,34 @@ def likeliest_error_sat_problem(
     flip any logical observable while producing no detection events).
 
     There are many tools that can solve maxSAT problems in WDIMACS format.
-    For example, you can download one of the entries in the 2023 maxSAT
-    competition (see https://maxsat-evaluations.github.io/2023), for example
-    CASHWMaxSAT-CorePlus.zip and run it on your problem by running these
-    BASH terminal commands:
+    One quick way to get started is to install pysat by running this BASH
+    terminal command:
+
+        pip install python-sat
+
+    Afterwards, you can run the included maxSAT solver "RC2" with this
+    Python code:
+
+        from pysat.examples.rc2 import RC2
+        from pysat.formula import WCNF
+
+        wcnf = WCNF(from_string="p wcnf 1 2 3\n3 -1 0\n3 1 0\n")
+
+        with RC2(wcnf) as rc2:
+        print(rc2.compute())  
+        print(rc2.cost)
+
+    Much faster solvers than RC2 are available on the 2023 maxSAT competition
+    website (see https://maxsat-evaluations.github.io/2023), for example
+    CASHWMaxSAT-CorePlus.zip. If you download this solver you can run it on your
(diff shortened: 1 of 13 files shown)


src/stim/search/sat/wcnf.cc
```diff
@@ -0,0 +1,229 @@
...
+        for (const auto &t : e.target_data) {
+            if (t.is_relative_detector_id()) {
+              detectors_activated[t.val()] = instance.Xor(detectors_activated[t.val()], err_x);
+            } else if (t.is_observable_id()) {
+              observables_flipped[t.val()] = instance.Xor(observables_flipped[t.val()], err_x);
+            }
+        }
+        // Add a soft clause for this error to be inactive
+        Clause clause;
+        clause.add_var(~err_x);
```
oscarhiggott, line 205: Since p>0.5 will lead to a disallowed negative weight, it might be worth handling the p>0.5 case with `clause.add_var(err_x);` and then flipping the sign of the weight?

src/stim/circuit/circuit.pybind.cc
```diff
@@ -1966,6 +1971,45 @@ void stim_pybind::pybind_circuit_methods(pybind11::module &, pybind11::class_<Ci
...
+                A WCNF file in [WDIMACS format](http://www.maxhs.org/docs/wdimacs.html)
+
+            Examples:
+                >>> import stim
+                >>> circuit = stim.Circuit.generated(
+                ...     "surface_code:rotated_memory_x",
+                ...     rounds=5,
+                ...     distance=5,
+                ...     after_clifford_depolarization=0.001)
+                >>> print(circuit.shortest_undetectable_logical_error_wcnf(
```
Strilanc, line 2005: Use a smaller example, so that you can include the exact output.

src/stim/circuit/circuit.pybind.cc
```diff
@@ -1966,6 +1971,45 @@ void stim_pybind::pybind_circuit_methods(pybind11::module &, pybind11::class_<Ci
...
+    c.def(
+        "shortest_undetectable_logical_error_wcnf",
+        &py_shortest_undetectable_logical_error_wcnf,
+        pybind11::kw_only(),
+        pybind11::arg("num_distinct_weights") = 1,
+        clean_doc_string(R"DOC(
+            Generates a maxSAT problem instance in WDIMACS format whose optimal value is
+            the distance of the protocol, i.e. the minimum weight of any set of errors
+            that forms an undetectable logical error.
+
```
Strilanc, line 1984: Include some notes on how you can use a publicly available solver to solve the result, like what you included in the PR message but shorter.

src/stim/search/sat/wcnf.test.cc
```diff
@@ -0,0 +1,55 @@
...
+  // hard clause x != (0, 1, 0)
+  // hard clause x != (1, 0, 0)
+  // hard clause x != (1, 1, 1)
+  // Plus 1 hard clause to ensure detector is not flipped
+  // hard clause -x_2
+  // Plus 1 hard clause to ensure an observable is flipped:
+  // hard clause x_0
+  // This gives a total of 8 clauses
+  // The top value should be at least 1 + 1 + 1 = 3. In our implementation ends up being 9.
+  std::stringstream expected;
```
Strilanc, line 36: Use a multiline string literal

```
expected = R"DIMACS(
so much space for activities!
)DIMACS";
```

src/stim/search/sat/wcnf.test.cc
```diff
@@ -0,0 +1,55 @@
+#include "stim/search/sat/wcnf.h"
+#include "gtest/gtest.h"
+
+using namespace stim;
+
+TEST(shortest_undetectable_logical_error_wcnf, no_error) {
+    // No error.
+    ASSERT_THROW(
```
Strilanc, line 8: Is this the right behavior for the empty case? Shouldn't it return a file that embeds an unsatisfiable problem instead of failing, in this case?

Failing fast is great. I like failing fast. But I also like my code to work on the empty string when its semantics are compatible with the method. I'd be quite cross if `''.startswith('')` raised an exception that I was probably doing the wrong thing, instead of returning true.

Strilanc (commented):
Looks like you need to run the `regen_docs.sh` script in the `dev/` directory. It expects you to build and install the stim dev wheel before doing so (see the developer docs; this is easy to do with bazel).

src/stim/search/sat/wcnf.cc
oscarhiggott: On second thoughts it may be best to leave it as is, since by default (unweighted case) the method finds the _shortest_ logical error, not the most probable.

noajshu: Hmm, good catch!
I agree with both your first and second thoughts and I am troubled by the potential for accidental mistakes in using this.

It's unclear to me what the best choice of API is that balances utility with safety.

Do you think it might be appropriate to throw an InvalidArgument or cerr a warning if there is any error mechanism with probability >= 0.5? This way it's at least a visible error. WDYT?

oscarhiggott: Yeah agree that just throwing an InvalidArgument about requiring p<0.5 is probably safest (and tbf it actually already does this with the message `"Clauses must have positive weight or HARD_CLAUSE_WEIGHT."` from line 58)

src/stim/search/sat/wcnf.test.cc
noajshu: We can certainly return an unsatisfiable wcnf in these cases.

src/stim/search/sat/wcnf.cc
noajshu: discussed offline and added another argument `weighted=False` where if it's True, does the p > 0.5 accounting Oscar suggested above.

src/stim/circuit/circuit.pybind.cc
```diff
@@ -1966,6 +1971,45 @@ void stim_pybind::pybind_circuit_methods(pybind11::module &, pybind11::class_<Ci
                 5
         )DOC")
             .data());
+
+    c.def(
+        "shortest_undetectable_logical_error_wcnf",
```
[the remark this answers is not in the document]
noajshu: Done

src/stim/circuit/circuit.pybind.cc
noajshu: Done

src/stim/circuit/circuit.pybind.cc
noajshu: Done

noajshu:
> Looks like you need to run the `regen_docs.sh` script in the `dev/` directory. It expects you to build and install the stim dev wheel before doing so (see the developer docs; this is easy to do with bazel).

Thanks, ran this and addressed all the comments.

Strilanc:
Note: the bug in `doctest_proper` is intended-behavior in `doctest`, where you have to say `<BLANKLINE>` instead of having blank output lines. But I prefer the modified print statements.

doc/python_api_reference_vDev.md
```diff
@@ -2437,6 +2438,77 @@ def search_for_undetectable_logical_errors(
...
+# stim.Circuit.shortest_error_problem_as_wcnf_file
+
+# (in class stim.Circuit)
+def shortest_error_problem_as_wcnf_file(
+    self,
+    *,
+    weighted: bool = False,
+    weight_scale_factor: int = 0,
+) -> str:
+    """Generates a maxSAT problem instance in WDIMACS format whose optimal value is
```
Strilanc, line 2452: Stim's docstrings follow the convention that the first line is a self-contained paragraph summarizing the method.

I've made an attempt at improving the docstring:

```
Makes a maxSAT problem of the circuit's distance, that other tools can solve.

The output is a string describing the maxSAT problem in WDIMACS format
(see https://maxhs.org/docs/wdimacs.html). The optimal solution to the
problem is the fault distance of the circuit (the minimum number of error
mechanisms that combine to flip any logical observable while producing no
detection events).

There are many tools that can solve maxSAT problems in WDIMACS format.
For example, you can download one of the entries in the 2023 maxSAT
competition (see https://maxsat-evaluations.github.io/2023) and run it on
your problem by running these BASH terminal commands:

    wget https://maxsat-evaluations.github.io/2023/mse23-solver-src/exact/CASHWMaxSAT-CorePlus.zip
    unzip CASHWMaxSAT-CorePlus.zip
    ./CASHWMaxSAT-CorePlus/bin/cashwmaxsatcoreplus -bm -m your_problem.wcnf

Args:
    weight_bins: Defaults to 1 (unweighted errors). Error probabilities are converted to
        log-odds and scaled/rounded to be positive integers at most this large. Setting
        this argument to a larger number results in the solution to the problem being the
        most likely logical error weighted by probability of its parts, instead of just the
        one with the fewest parts.

Returns:
    A string corresponding to the contents of a WCNF file in WDIMACS format.

Examples:
    ...
```

doc/python_api_reference_vDev.md
```diff
@@ -2437,6 +2438,77 @@ def search_for_undetectable_logical_errors(
...
+        p wcnf 2 4 4001
+        185 -1 0
+        1000 -2 0
+        4001 -1 0
+        4001 2 0
+    """
+```
+
 <a name="stim.Circuit.shortest_graphlike_error"></a>
 ```python
```
Strilanc, line 2513: This method should also be on `DetectorErrorModel`

src/stim/search/sat/wcnf.cc
```diff
@@ -0,0 +1,261 @@
...
+  }
+  if (!weighted and weight_scale_factor != 0) {
+    throw std::invalid_argument("for unweighted problems, weight_scale_factor must be == 0");
+  }
+
+  size_t num_observables = model.count_observables();
+  size_t num_detectors = model.count_detectors();
+  size_t num_errors = model.count_errors();
+  if (num_observables == 0 or num_detectors == 0 or num_errors == 0) {
+    return UNSAT_WCNF_STR;
```
Strilanc, line 186: This is wrong. The no-det yes-err yes-obs case is often solvable with a result of d=1.

Add unit tests including a circuit with obs but no dets, dets but no obs, no obs and no dets, no errors, etc. Hit the corner cases.

Ideally, the method should not really check for the no-obs case, and instead just do what it normally does resulting in an unsatisfiable problem.

src/stim/search/sat/wcnf.cc
```diff
@@ -0,0 +1,261 @@
...
+    if (weight == HARD_CLAUSE_WEIGHT) {
+      // Hard clause
+      return top;
+    }
+    // Soft clause
+    if (!weighted) {
+      // For unweighted problems, all soft clauses have weight 1.
+      return 1;
+    }
+    return std::round(weight / max_weight * (double)weight_scale_factor);
```
Strilanc, line 127: This should be rounding such that it's never 0, right? It looks like this might end up as 0.

Add a unit test with an error of probability 49% as well as an error of probability 0.00000000001%.

src/stim/search/sat/wcnf.cc
```diff
@@ -0,0 +1,261 @@
...
+        if (var.negated) {
+          ss << " -" << (var.variable + 1);
+        } else {
+          ss << " " << (var.variable + 1);
+        }
+      }
+      // Each clause ends with 0
+      ss << " 0\n";
+    }
+    return ss.str();
```
Strilanc, line 167: Nit: I'd refactor the method into a `write_dimacs` style, taking a `std::ostream &out` parameter, instead of forcing the allocation of the string. Not necessary.

src/stim/search/sat/wcnf.cc
```diff
@@ -0,0 +1,261 @@
...
+          clause.weight = -std::log(p / (1 - p));
+          instance.add_clause(clause);
+        }
+      }
+      ++error_index;
+  });
+  assert(error_index == num_errors);
+
+  // Add a hard clause for each detector to be inactive
+  for (size_t d=0; d<num_detectors; ++d) {
```
Strilanc, line 246: Nit: Run

```
find src | grep "\.\(cc\|h\|inl\)$" | grep -Pv "crumble_data.cc|gate_data_3d_texture_data.cc" | xargs clang-format -i
```

to autoformat the code.

src/stim/search/sat/wcnf.test.cc
Strilanc: I don't understand why tabs in the data would prevent the use of multiline string literals.

Note that, for multiline literals, you don't indent the contents of the literal.

Strilanc (changes requested):
Getting closer!

src/stim/search/sat/wcnf.cc
Strilanc: The docstring should probably explicitly say how p>0.5 errors are inverted if `weight_bins > 1`.

noajshu:
Hi @Strilanc , Oscar had an idea that I like for the interface and I was wondering what you thought.

The idea is to split into two python functions:

```python
minimum_cardinality_undetectable_logical_error_problem_as_maxsat_string(format)
approximately_most_probable_undetectable_logical_error_problem_as_maxsat_string(quantization_factor, format)
```

The idea would be to to ignore probabilities entirely in the first method and do a "best effort" approximation in the second method. "Best effort" means negation complementation and rounding to zero where it's the closest available integer.

WDYT

Strilanc:
I think these are close enough that they might as well be the same method.

Also those names are too long.

noajshu:
> I think these are close enough that they might as well be the same method with a flag of some kind.
> 
> Also those names are too long. If I was naming it I'd say `shortest_logical_error_as_...` and `most_likely_logical_error_as_...`.

I am happy to make either 2 methods with somewhat shorter names, as you suggest, or to keep it 1 method with a flag. Which solution do you prefer?

Strilanc:
Go with the two name solution.

src/stim/search/sat/wcnf.test.cc
noajshu: I'm going to push back against this suggestion again because it prevents adding explanatory inline comments between each of the lines.

doc/python_api_reference_vDev.md
noajshu: That sounds nice, but could we leave that for a future PR after we've solidified the API here?
Also, note that the heuristic approach `search_for_undetectable_logical_errors` is not a method of `DetectorErrorModel`. It seems like if we add the SAT method to `DetectorErrorModel` then why not also add the heuristic method.
WDYT?

doc/python_api_reference_vDev.md
noajshu: Thanks. I had actually omitted the url because I got this error:
```
E   ImportError: Docstring line has length 98 > 80:
E       wget https://maxsat-evaluations.github.io/2023/mse23-solver-src/exact/CASHWMaxSAT-CorePlus.zip
E   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Any idea how to get past it?

doc/python_api_reference_vDev.md
```diff
@@ -46,6 +46,7 @@ API references for stable versions are kept on the [stim github wiki](https://gi
     - [`stim.Circuit.num_ticks`](#stim.Circuit.num_ticks)
     - [`stim.Circuit.reference_sample`](#stim.Circuit.reference_sample)
     - [`stim.Circuit.search_for_undetectable_logical_errors`](#stim.Circuit.search_for_undetectable_logical_errors)
+    - [`stim.Circuit.shortest_error_problem_as_wcnf_file`](#stim.Circuit.shortest_error_problem_as_wcnf_file)
```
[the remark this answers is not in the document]
noajshu: done

doc/python_api_reference_vDev.md
```diff
@@ -2437,6 +2438,77 @@ def search_for_undetectable_logical_errors(
...
+    # https://maxsat-evaluations.github.io/2023
+    unzip CASHWMaxSAT-CorePlus.zip
+    time ./CASHWMaxSAT-CorePlus/bin/cashwmaxsatcoreplus -bm -m problem.wcnf 
+    ```
+
+    Args:
+        weighted: Defaults to False (unweighted), so that the problem is to find
+        the minimum size set of errors that leads to an undetectable logical
+        error.
+        weight_scale_factor: Defaults to 0. If weighted==False, must set
```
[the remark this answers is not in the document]
noajshu: discussed offline, we now have 2 methods so this is out of date

src/stim/search/sat/wcnf.cc
noajshu: > This is wrong. The no-det yes-err yes-obs case is often solvable with a result of d=1.

good catch, fixed!

> Add unit tests including a circuit with obs but no dets, dets but no obs, no obs and no dets, no errors, etc. Hit the corner cases.

Done

> Ideally, the method should not really check for the no-obs case, and instead just do what it normally does resulting in an unsatisfiable problem.

I like your suggestion to 'do what it normally does'. Unfortunately SAT formulas do not permit clauses that involve constants which makes translating this edge case a bit unnatural to handle later in the translation.

If you would like aspects of the implementation to be changed I suggest we defer that to a future PR once the API and tests are solidified here.

src/stim/search/sat/wcnf.cc
noajshu: I added these unit tests for the shortest error problem but for the likeliest error problem it's really supposed to round down to 0 in this case.

src/stim/search/sat/wcnf.cc
noajshu: thanks! done

src/stim/search/sat/wcnf.cc
noajshu: Ok, thanks.

src/stim/search/sat/wcnf.cc
Strilanc: Won't that result in instances that appear to be unsatisfiable, despite having a solution that's merely really-unlikely?

doc/python_api_reference_vDev.md
```diff
@@ -2437,6 +2438,77 @@ def search_for_undetectable_logical_errors(
...
+    ```
+    # first download nzip CASHWMaxSAT-CorePlus.zip from 
+    # https://maxsat-evaluations.github.io/2023
+    unzip CASHWMaxSAT-CorePlus.zip
+    time ./CASHWMaxSAT-CorePlus/bin/cashwmaxsatcoreplus -bm -m problem.wcnf 
+    ```
+
+    Args:
+        weighted: Defaults to False (unweighted), so that the problem is to find
+        the minimum size set of errors that leads to an undetectable logical
```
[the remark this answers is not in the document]
noajshu: done

doc/python_api_reference_vDev.md
Strilanc: Go to https://github.com/quantumlib/Stim/blob/1cdf45f884442d01806d6a8997b867a352e937c9/src/stim/cmd/command_help.cc#L74 and add `"wget "` as a prefix string that disables the check. Or change it to allow if `https://` appears anywhere instead of just at the start.

src/stim/search/sat/wcnf.cc
noajshu: Done

doc/python_api_reference_vDev.md
noajshu: Thank you!!

noajshu: That doesn't work. I still get:
```
E   ImportError: Docstring line has length 98 > 80:
E       wget https://maxsat-evaluations.github.io/2023/mse23-solver-src/exact/CASHWMaxSAT-CorePlus.zip
E   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

noajshu: I think the error is happening later on when we try to import the module, not when we clean the docstring

noajshu: Ah nvm, I got it 🤦

noajshu:
@Strilanc despite running this:
```
bazel build stim_dev_wheel
pip uninstall -y stim
pip install bazel-bin/stim-0.0.dev0-py3-none-any.whl
pytest src/stim/circuit/ && ./dev/regen_docs.sh
```
I don't see any updates to the docs. Yet, the CI job `test_generated_docs_are_fresh` does not pass. Do you have any idea what could be going wrong?

Strilanc:
The failure message i nthe CI log is

image

So it looks like your doctest code is invalid. The fresh-check also verifies the .pyi file runs as if it was python. You just need to make it be exactly valid (need '...' on every line including in multiline strings)

noajshu:
Thanks for the advice!
The problem with the docstring was simply that I used:
```
>>> circuit = stim.Circuit("""
...   X_ERROR(0.1) 0
...   M 0
...   OBSERVABLE_INCLUDE(0) rec[-1]
...   X_ERROR(0.4) 0
...   M 0
...   DETECTOR rec[-1] rec[-2]
... """)
```
which resulted in a """ inside a multiline string that itself used """.

I replaced these with ''' and also added a check to `clean_doc_string` to raise an error if there are any appearances of """ to help future fledgling Stim initiates like myself.

BTW -- it seems like some of the code in a recent commit to main was not formatted with this command you shared:
```
find src | grep "\.\(cc\|h\|inl\)$" | grep -Pv "crumble_data.cc|gate_data_3d_texture_data.cc" | xargs clang-format -i
```
you might consider running it on your branch.

Strilanc:
Because CI doesn't enforce that the code is formatted, it often drifts a bit over time.

Strilanc (approved):

viathor:
Cross-reference: This PR came up as an example to answer [this question](https://quantumcomputing.stackexchange.com/q/37289/10480) on QCSE.

PR #718, opened 2024-03-15
Debug address sanitization crashes in github actions

Strilanc:
Trying to figure out where the test address sanitization errors are coming from. Can't repro locally so doing it by pushing to this branch to trigger github actions.


Strilanc:
Experimentation complete, workaround identified.

Strilanc:
Oh, I see that there's like no granularity in choosing images so the workaround of using ubuntu-20 doesn't actually work when using C++20...

Issue #717, opened 2024-03-15
Figure out why `test (256)` keeps failing


Strilanc:
Sometimes (only sometimes!) it gets stuck in an infinite loop of printing out address sanitizer errors. Very concerning, but only seems to happen for the 256-avx case so far. This one is currently not allowed to go into the python wheels due to a previous mysterious crash, so this isn't urgent, but it sure would be nice to know why these things are occurring.

![image](https://github.com/quantumlib/Stim/assets/79941/494b4b90-b54b-4827-a236-00755effa09c)

Strilanc:
Nevermind it has now also happened for `test (64)` and is therefore very concerning.

![image](https://github.com/quantumlib/Stim/assets/79941/51c2c381-8f27-4eb0-a4eb-9f0db2d96ff2)

Strilanc:
Based on https://github.com/quantumlib/Stim/pull/718 this is a bug in gtest rather than a bug in stim.

kimwalisch:
I have the same bug in my [primecount](https://github.com/kimwalisch/primecount) project in the code below which only uses the math functions from the C++ standard library:

```C++
  for (int i = 0; i < 100; i++)
  {
    T term = (Li(t) - x) * std::log(t);

    // Not converging anymore
    if (std::abs(term) >= std::abs(old_term))
      break;

    t -= term;
    old_term = term;
  }
```

My bug only occurs on Ubuntu 22.04 & 23.10 when running in a virtual machine and enabling the GCC/Clang sanitizers. When I switched my CI test to ubuntu-20.04 the bug disappeared. (When I tested using Ubuntu 22.04 & GCC sanitizers on a real server (no VM or Docker container) it also works without any issues)

After more than 2 hours of debugging I couldn't figure out the exact cause of the issue, but it looks like the issue is caused by a Ubuntu >= 22.04 bug or a compiler/sanitizer bug.

Strilanc:
@kimwalisch Thanks, that's very helpful to know that I can work around it by pinning the version of ubuntu used by CI.

Strilanc:
This seems to have been resolved externally. Hasn't happened in a PR for about a week now, whereas before it was happening multiple times per PR.

PR #719, opened 2024-03-17
Work around false CI failures by pinning to ubuntu-20

Strilanc:
Fixes https://github.com/quantumlib/Stim/issues/717 (whatever ubuntu-latest bug is causing it is not fixed, but for continuing-to-develop-Stim purposes it is fixed)


Strilanc:
Gah, doesn't support C++20

PR #728, opened 2024-03-21
Fix cirq-to-stim conversion not adding `TICK`s for multi-moment single-rep `CircuitOperation`s

AlexBourassa:
Currently in `main` when we convert a cirq circuit to stim with a `CircuitOperation` with a single repetition (or with `flatten=True`) no TICK instructions are inserted since the moments are removed in the `cirq.decompose_once(op)` command.

### DIFF
--- a/glue/cirq/stimcirq/_cirq_to_stim_test.py
+++ b/glue/cirq/stimcirq/_cirq_to_stim_test.py
@@ -339,27 +339,30 @@ def test_on_loop():
     result = stimcirq.StimSampler().run(c)
     assert result.measurements.keys() == {'0:a', '0:b', '1:a', '1:b', '2:a', '2:b'}
 
-def test_loop_with_one_rep():
-    q0 = cirq.q(0)
+
+def test_multi_moment_circuit_operation():
+    q0 = cirq.LineQubit(0)
     cc = cirq.Circuit(
         cirq.CircuitOperation(
-        cirq.FrozenCircuit(
-            cirq.Moment(cirq.H(q0)),
-            cirq.Moment(cirq.H(q0)),
-            cirq.Moment(cirq.H(q0)),
-            cirq.Moment(cirq.H(q0)),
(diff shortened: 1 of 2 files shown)


glue/cirq/stimcirq/_cirq_to_stim_test.py
```diff
@@ -339,6 +339,21 @@ def test_on_loop():
...
+        cirq.FrozenCircuit(
+            cirq.Moment(cirq.H(q0)),
+            cirq.Moment(cirq.H(q0)),
+            cirq.Moment(cirq.H(q0)),
+            cirq.Moment(cirq.H(q0)),
+        )
+        )
+    )
+    sc = stimcirq.cirq_circuit_to_stim_circuit(cc)
+    assert str(sc).splitlines().count('TICK') > 1
```
Strilanc, line 355: Assert the actual circuit's contents instead of asserting something abstract about it.

Strilanc:
Can you give an example of a current bad output in the description of the PR?

glue/cirq/stimcirq/_cirq_to_stim_test.py
AlexBourassa: Made the change!

AlexBourassa:
> Can you give an example of a current bad output in the description of the PR?

On `main`, the stim circuit from the unit test would be:
```
H 0 0 0 0
TICK
```

Whereas after this PR the output circuit is
```
H 0
TICK
H 0
TICK
H 0
TICK
H 0
TICK
```

AlexBourassa:
@Strilanc I think this should be ready to go!  (although some of the CI are failing, but I don't think this is related to my change?).  Please take another look when you get a chance!

Strilanc:
It was failing because you were using `cirq.q` which doesn't exist in the cirq version the tests install, to ensure backwards compatibility.

PR #732, opened 2024-03-27
Adding support to not native x86 systems for external libraries

mghibaudi:
- Filtering of x86 specific directive at both `Makefile` and `setup.py` level. Former needed for libraries that integrate Stim at the binary level (e.g. PyMatching).
- Replacing `sys` with `platform` to gain access to additional information on the OS and CPU architecture. 
- Wrapped up with a function the extensions that shall be built for a specific architecture.


Strilanc:
Is there a way to test these working, from github actions, so that I don't accidentally break them given that I'm running on x86 machines?

Strilanc:
The new muxing logic is failing on widows and osx, as you can see in the CI tests.

mghibaudi:
Thanks @Strilanc for the quick review - enabled CI on the fork and debugged the issues (i386, i686 were missing - thus Win and Mac failures). Current version passes cleanly on my forkd' main.

mghibaudi:
@Strilanc - just checking did you manage to have a look at the latest changes? they do fix the previous error and pass in CI (on the forked branch).

PR #734, opened 2024-03-28
Fix auto dem condition flow

inmzhang:
In my understanding, the condition flow in the `sinter.WorkIn.auto_dem()` should be: 

`(decompose_errors=True, flatten_loops=False)` -> `(decompose_errors=True, flatten_loops=True)` ->`(decompose_errors=False, flatten_loops=False)` ->`(decompose_errors=False, flatten_loops=True)`

while the current code tried `(decompose_errors=True, flatten_loops=False)` and `(decompose_errors=False, flatten_loops=False)` twice, which seems like typos. Did I miss something?


Strilanc:
Urgh that's embarrassing that I got the order wrong. But I think this PR still has the wrong order. It should be

```
{flatten,decompose}
{decompose}
{flatten}
{}
```

The reason for flatten+decompose first is because flattening makes the conversion faster and the dem's file size smaller. There's a bug where it can cause the decomposition to fail, so there's the fallback to no-flattening. Probably this should just be changed to a fallback to no-flattening yes-decompose with ignore_decomposition_failures=True.

I'll think about it some more.

Strilanc:
Moved this change into https://github.com/quantumlib/Stim/pull/832

PR #735, opened 2024-03-28
Sinter custom sampler

### LINKED ISSUE #682: Sinter Custom Sampler
Similar to how sinter supports custom decoders, it should support custom samplers.

Use cases for this include someone lightly wrapping the flip or tableau simulator, or using a more complex tool like a cirq or numerical simulator to produce the measurement or detector outcomes. 

This should allow people to avoid re-implementing sinter's nice multiprocessing approach to sampling when they want to make a change to how the sampler works.

inmzhang:
Fixes #682

This issue is related to one of my projects, so I attempted a basic and quick solution. I followed the pattern of `sinter.Decoder` and the implementation between the two should be similar. 

The sampler has been poorly tested, and there may still be some issues with backward compatibility. However, I would appreciate your initial thoughts, as I'm not sure if this aligns with your design for this issue.


glue/sample/src/sinter/_collection_work_manager.py
```diff
@@ -255,16 +276,25 @@ def _iter_tasks_with_assigned_decoders(
                 circuit_path=task.circuit_path,
             )
 
+        if default_samplers is not None:
+            task_samplers = list(default_samplers)
+        elif task.sampler is not None:
+            task_samplers = [task.sampler]
```
Strilanc, line 282: This is the wrong order. Specific should override default (or append to it).

Strilanc (changes requested):
This looks pretty good! I'll try it out for a bit before approving since it's a big change.

Sorry I took so long to review it. I think the main missing thing is to document it by editing the README and the getting started notebook. You may want to add an example of using it to the getting started notebook as well.

glue/sample/src/sinter/_sampling_sampler_class.py
```diff
@@ -0,0 +1,123 @@
...
+            the preconfigured sampler.
+
+        Raises:
+            NotImplementedError: This `sinter.Sampler` doesn't support compiling
+                for a circuit.
+        """
+        raise NotImplementedError("compile_sampler_for_circuit")
+
+    @abc.abstractmethod
+    def sample_detectors_via_files(
```
Strilanc, line 86: Drop this method. The files thing was a mistep; no need to repeat it.

glue/sample/src/sinter/_sampling_sampler_class.py
```diff
@@ -0,0 +1,123 @@
...
+
+    This is the type returned by `sinter.Sampler.compile_sampler_for_circuit`. The
+    idea is that, when many shots of the same sampling task are going to be
+    performed, it is valuable to pay the cost of configuring the sampler only
+    once instead of once per batch of shots. Custom samplers can optionally
+    implement that method, and return this type, to increase sampling
+    efficiency.
+    """
+    @abc.abstractmethod
+    def sample_detectors_bit_packed(
```
Strilanc, line 20: I think, instead of sampling detection events, it would be better to sample `AnonTaskStats` directly. This would allow for much more flexibility in what a sampler did. But then I'm not sure how the decoder exactly fits into the picture... is it an argument to this method? Bleh...

inmzhang: Thanks for the review! Could you give a usecase where sampling `AnnoTaskStats` will help? I'm not sure how to settle down the interface yet since it will introduce strong coupling between `Sampler` and `Decoder` if I feed the decoder as an argument to the sampling method.

Strilanc:
Sorry for not responding to this for awhile. I took some inspiration from this PR and started iterating on it, and ultimately ended up going in a different direction. Specifically, I added a `sinter.Sampler` class and allowed it to go almost anywhere a `sinter.Decoder` can currently go, with the understanding that the sampler will take full responsibility for generating samples and producing statistics instead of only handling the decoding. This made fewer changes to the API surface, and fixed a nasty annoyance where not every sampler is compatible with every decoder (or even needs a decoder) but sinter was still asking for a decoder.

Closing as obsoleted by https://github.com/quantumlib/Stim/pull/804

PR #740, opened 2024-04-08
Pytype fixes

folded:
Ensure that python code bundled with stim passes `pytype --strict-none-binding`.

### DIFF
--- a/glue/sample/src/sinter/_collection.py
+++ b/glue/sample/src/sinter/_collection.py
@@ -2,7 +2,7 @@ import contextlib
 import dataclasses
 import pathlib
 from typing import Any
-from typing import Callable, Iterator, Optional, Union, Iterable, List, TYPE_CHECKING, Tuple, Dict
+from typing import Callable, Iterator, Optional, Union, Iterable, List, Tuple, Dict
 
 import math
 import numpy as np
@@ -11,12 +11,11 @@ import stim
 from sinter._collection_options import CollectionOptions
 from sinter._csv_out import CSV_HEADER
 from sinter._collection_work_manager import CollectionWorkManager
+from sinter._decoding_decoder_class import Decoder
 from sinter._existing_data import ExistingData
 from sinter._printer import ThrottledProgressPrinter
+from sinter._task import Task
 from sinter._task_stats import TaskStats
-
-if TYPE_CHECKING:
-    import sinter
 
 
 @dataclasses.dataclass(frozen=True)
@@ -39,8 +38,8 @@ class Progress:
 
 def iter_collect(*,
                  num_workers: int,
-                 tasks: Union[Iterator['sinter.Task'],
-                              Iterable['sinter.Task']],
+                 tasks: Union[Iterator[Task],
+                              Iterable[Task]],
                  hint_num_tasks: Optional[int] = None,
                  additional_existing_data: Optional[ExistingData] = None,
                  max_shots: Optional[int] = None,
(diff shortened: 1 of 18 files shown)


glue/sample/src/sinter/_collection.py
```diff
@@ -220,10 +219,10 @@ def iter_collect(*,
 
 def collect(*,
             num_workers: int,
-            tasks: Union[Iterator['sinter.Task'], Iterable['sinter.Task']],
+            tasks: Union[Iterator[Task], Iterable[Task]],
```
Strilanc, line 222: Is this necessary to change? Because these methods are in the external API, I prefer to name the types how users would use them ("sinter.Task" instead of just "Task").

glue/sample/src/sinter/_decoding_pymatching.py
```diff
@@ -10,12 +19,15 @@ def decode_shots_bit_packed(
...
     ) -> 'np.ndarray':
-        return self.matcher.decode_batch(
+        result = self.matcher.decode_batch(
             shots=bit_packed_detection_event_data,
             bit_packed_shots=True,
             bit_packed_predictions=True,
             return_weights=False,
         )
+        if isinstance(result, tuple):
+          return result[0]  # pymatching returned predictions and weights
```
Strilanc, line 29: If this ever actually executed, it would be a bug. Raise an exception, or do a cast.

glue/sample/src/sinter/_decoding_test.py
```diff
@@ -361,7 +361,7 @@ def test_decode_fails_correctly(decoder: str, force_streaming: Optional[bool]):
         with open(d / 'bad_dets.b8', 'wb') as f:
             f.write(b'!')
 
-        if 'vacuous' not in decoder:
+        if 'vacuous' not in decoder and decoder_obj is not None:
```
Strilanc, line 364: I think this should assert it's not none instead of disabling the testing of the decoder that wasn't.

glue/sample/src/sinter/_plotting.py
```diff
@@ -337,12 +350,12 @@ def _rescale(v: np.ndarray, scale: str, invert: bool) -> np.ndarray:
...
-        y_func: Callable[['sinter.TaskStats'], Union['sinter.Fit', float, int]],
-        group_func: Callable[['sinter.TaskStats'], TCurveId] = lambda _: None,
-        filter_func: Callable[['sinter.TaskStats'], Any] = lambda _: True,
-        plot_args_func: Callable[[int, TCurveId, List['sinter.TaskStats']], Dict[str, Any]] = lambda index, group_key, group_stats: dict(),
+        stats: Iterable[TaskStats],
+        x_func: Callable[[TaskStats], Any],
+        y_func: Callable[[TaskStats], Union[Fit, float, int]],
+        group_func: _GroupFunc[TCurveId] = lambda _: None,
+        filter_func: Callable[[TaskStats], Any] = lambda _: True,
+        plot_args_func: Union[_PlotArgsFunc2[TCurveId], _PlotArgsFunc3[TCurveId]] = lambda index, group_key, group_stats: dict(),
```
Strilanc, line 358: This is an example where I sorta prefer for the exposed API to not mention the deprecated functionality, even though the method still technically supports it. I can't remove the backwards compat, but I can hide the complication of the backwards compat from new users.

Strilanc:
Closing due to stale comments and growing merge conflicts. Still open to fixing type errors in downstream tools.

PR #751, opened 2024-04-12
Resolved ambiguity in simd_word test by explicitly casting size_t to uint64_t for << operator usage.

sengthai:


Strilanc:
What is the ambiguity that's being resolved?

sengthai:
> What is the ambiguity that's being resolved?

The ambiguity is when using the `<<` operator between `simd_word<W>` and `size_t` type. The compiler may not determine which overloaded `<<` operator to use. By casting one operand to `uint64_t`, we specify which operator to use, eliminating the ambiguity.

AntonMenshov:
I confirm that without this fix, the compilation on AppleClang 15.0.0 is problematic (essentially, boiling down to the compiler erroring out on trying to find the right match for the `<<` operator in the aforementioned unit test).

sengthai:
@AntonMenshov  Oh I see. Thank you for confirmation.

Strilanc:
Sorry for the delay on merging this. I see you closed it; I moved the changes over to https://github.com/quantumlib/Stim/pull/758

PR #777, opened 2024-06-06
Change return type of CircuitRepeatBlock.name to str

### LINKED ISSUE #776: Small typing issue in the Python method `CircuitRepeatBlock.name`
The method below

https://github.com/quantumlib/Stim/blob/985d6a7042229827473d620d539119e3dbe0946e/glue/python/src/stim/__init__.pyi#L2840-L2861

is guaranteed to return `"REPEAT"` per the docstring. Nevertheless, its annotated return type is `object`, that is not invalid per-se but is not as precise as it should be (i.e., `str`).

nelimee:
The correct return type for this method should be `str` as the documentation guarantees that this specific method returns `"REPEAT"`.

Fixes #776

### DIFF
--- a/glue/python/src/stim/__init__.pyi
+++ b/glue/python/src/stim/__init__.pyi
@@ -2840,7 +2840,7 @@ class CircuitRepeatBlock:
     @property
     def name(
         self,
-    ) -> object:
+    ) -> str:
         """Returns the name "REPEAT".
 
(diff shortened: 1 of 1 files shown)


Strilanc:
Good idea, but you edited the generated docs instead of the source. I finished this in https://github.com/quantumlib/Stim/pull/783

PR #786, opened 2024-06-21
Fix order of operations in Tableau.__call__.__doc__

amirebrahimi:
This is minor, but I figure it might save someone else some time.

### DIFF
--- a/doc/python_api_reference_vDev.md
+++ b/doc/python_api_reference_vDev.md
@@ -10077,9 +10077,8 @@ def __call__(
 ) -> stim.PauliString:
     """Returns the conjugation of a PauliString by the Tableau's Clifford operation.
 
-    The conjugation of P by C is equal to C**-1 * P * C (circuit order where C**-1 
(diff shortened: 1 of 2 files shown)


amirebrahimi:
I just signed the CLA, so not sure if it will require CI to rerun.

Strilanc:
Everyone always edits the generated docs instead of the source.

If the multiplication is in circuit order, then that's the wrong way around. I should fix that and then also separately describe it as "before/after" so it's clearer.

amirebrahimi:
This seems to be consistent with what is specified [here](https://github.com/quantumlib/Stim/blob/main/src/stim/stabilizers/tableau.h#L133).

Strilanc:
Fixes https://github.com/quantumlib/Stim/issues/819

Strilanc:
Moved to https://github.com/quantumlib/Stim/pull/822 so I could edit it to get it merged. Thanks for the fix!

PR #790, opened 2024-07-02
Add .bazelversion

dougthor42:
This repo is already using `bazelisk` but does not currently pin a Bazel version.

Add `.bazelversion` to pin Bazel based on the version used in a recent CI run (https://github.com/quantumlib/Stim/actions/runs/9657364117/job/26636504001).

Related to https://github.com/qh-lab/pyle/pull/46833#issuecomment-2204296035

/cc @Strilanc


Strilanc:
SGTM

PR #815, opened 2024-08-24
add hypergraph-UF and mwpf decoder in sinter

yuewuo:
This is a simple adaptation of the `mwpf' python package to sinter. Just copied the `_decoding_fusion.py` and modify it accordingly to let the decoder consider hyperedges. Tested using the `getting_started.ipynb` and they run decoding without a problem.

I haven't added any test cases for this, since I didn't find a test case for `fusion_blossom` decoder as well. Please let me know where should I put a basic test case for the new decoders!

### DIFF
--- a/glue/sample/src/sinter/_decoding_all_built_in_decoders.py
+++ b/glue/sample/src/sinter/_decoding_all_built_in_decoders.py
@@ -10,6 +10,8 @@ BUILT_IN_DECODERS: Dict[str, Decoder] = {
     'vacuous': VacuousDecoder(),
     'pymatching': PyMatchingDecoder(),
     'fusion_blossom': FusionBlossomDecoder(),
-    'hyper_uf': HyperUFDecoder(),
-    'mwpf': MwpfDecoder(),
+    # an implementation of (weighted) hypergraph UF decoder (https://arxiv.org/abs/2103.08049)
+    'hypergraph_union_find': HyperUFDecoder(),
+    # Minimum-Weight Parity Factor using similar primal-dual method the blossom algorithm (https://pypi.org/project/mwpf/)
+    'mw_parity_factor': MwpfDecoder(),  
 }
--- a/glue/sample/src/sinter/_decoding_all_built_in_decoders.py
+++ b/glue/sample/src/sinter/_decoding_all_built_in_decoders.py
@@ -4,9 +4,12 @@ from sinter._decoding_decoder_class import Decoder
 from sinter._decoding_fusion_blossom import FusionBlossomDecoder
 from sinter._decoding_pymatching import PyMatchingDecoder
 from sinter._decoding_vacuous import VacuousDecoder
+from sinter._decoding_mwpf import HyperUFDecoder, MwpfDecoder
 
 BUILT_IN_DECODERS: Dict[str, Decoder] = {
     'vacuous': VacuousDecoder(),
     'pymatching': PyMatchingDecoder(),
     'fusion_blossom': FusionBlossomDecoder(),
+    'hyper_uf': HyperUFDecoder(),
+    'mwpf': MwpfDecoder(),
 }
--- a/glue/sample/src/sinter/_decoding_mwpf.py
+++ b/glue/sample/src/sinter/_decoding_mwpf.py
@@ -1,6 +1,6 @@
 import math
 import pathlib
-from typing import Callable, List, TYPE_CHECKING, Tuple
+from typing import Callable, List, TYPE_CHECKING, Tuple, Any
 
 import numpy as np
 import stim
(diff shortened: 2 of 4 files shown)


glue/sample/src/sinter/_decoding_all_built_in_decoders.py
```diff
@@ -4,9 +4,12 @@
...
 from sinter._decoding_pymatching import PyMatchingDecoder
 from sinter._decoding_vacuous import VacuousDecoder
+from sinter._decoding_mwpf import HyperUFDecoder, MwpfDecoder
 
 BUILT_IN_DECODERS: Dict[str, Decoder] = {
     'vacuous': VacuousDecoder(),
     'pymatching': PyMatchingDecoder(),
     'fusion_blossom': FusionBlossomDecoder(),
+    'hyper_uf': HyperUFDecoder(),
+    'mwpf': MwpfDecoder(),
```
Strilanc, line 14: Could these names be more descriptive?

glue/sample/src/sinter/_decoding_mwpf.py
```diff
@@ -0,0 +1,262 @@
...
+
+import numpy as np
+import stim
+
+from sinter._decoding_decoder_class import Decoder, CompiledDecoder
+
+if TYPE_CHECKING:
+    import mwpf
+
+DEFAULT_TIMEOUT: float = 10.0  # decoder timeout in seconds
```
Strilanc, line 13: Whoah whoah whoah, this timeout is a major problem.

Sinter is a bulk sampling library. The workers collecting samples could be executing under widely different loads, or on machines of widely different speeds. Specifying a timeout in seconds means that you could get different statistics from running on different machines, or at different times, or with different numbers of workers. It completely breaks the assumption that the collected samples can be merged while assuming that they are from the same distribution.

This needs to be replaced with something like a maximum number of steps, or a chance-of-giving-up-per-step, or some other quantity that will give the same distribution today as it would give tomorrow.

Back in the old days there were some compilers that used timeouts to determine when to stop optimizing a program. They were a disaster, because the performance of the resulting program would change from run to run.

Strilanc (changes requested):
Thanks for the contribution. Unfortunately, I think there's currently a bit of a showstopper. The timeout behavior needs to be replaced in order for results to be mergeable (and reproduceable) across machines.

glue/sample/src/sinter/_decoding_mwpf.py
yuewuo: Yeah, this makes a lot of sense! Thanks for the feedback and I'll add mechanism to deterministically limit the runtime. 

One question is whether it's possible to pass a configuration object to the decoder construction? I don't see such configuration in PyMatching or other decoders so I'm just wondering what's your suggestion of doing this, especially when everything is handled by sinter.Task. It could be useful for other decoders as well.

yuewuo:
A basic performance benchmark using the sinter jobs of `getting_started.ipynb` (the `sinter.collect(` block of `repetition_code:memory`)
- pymatching: 1.7s
- fusion_blossom: 35s
- hyper_uf: 89.5s
- mwpf: 142.9s

By the way, I had some thoughts on the slow down of `fusion_blossom` compared to `pymatching`: it is probably coming from duplicate edges. The current implementation of sinter for fusion_blossom does not deduplicate the edges (and merge them) like pymatching would do internally when calling the `add_edge` function. I think there could be some functionality outside of pymatching that merges the edges. Do you think this is reasonable?

glue/sample/src/sinter/_decoding_mwpf.py
yuewuo: I have updated the implementations to avoid timeout. Now the class corresponds to a decoder without any timeout, so it's probably running too long for some syndromes. We'll add some mechanism later, but the current implementation seems to work, with some benchmark results commented below.

Strilanc:
Having duplicate edges could certainly slow things down. Stim already dedupes most errors so there may not be a lot of dupes to start with.

yuewuo:
> Having duplicate edges could certainly slow things down. Stim already dedupes most errors so there may not be a lot of dupes to start with.

Thanks for the insight! Indeed, just printed them out and there are no duplicate edges. I've also tried to reduce the integer weight of the edges, but the decoding time remains almost unchanged. The most probably reason would then because iterating the ndarray in Python instead of in Rust. I'll investigate in this to get a better understanding.

yuewuo:
> > Having duplicate edges could certainly slow things down. Stim already dedupes most errors so there may not be a lot of dupes to start with.
> 
> Thanks for the insight! Indeed, just printed them out and there are no duplicate edges. I've also tried to reduce the integer weight of the edges, but the decoding time remains almost unchanged. The most probably reason would then because iterating the ndarray in Python instead of in Rust. I'll investigate in this to get a better understanding.

Sorry I made a stupid mistake...... I installed the debug version of the `fusion_blossom` package which runs for 35s. Once I switch to a release version, it drops to 6.3s. This makes more sense. Similarly for `mwpf` package. Here is the new result:

- pymatching: 2.3s
- fusion_blossom: 6.3s
- hypergraph_union_find: 9.7s
- mw_parity_factor: 14.0s

Strilanc (approved):

Strilanc:
I moved this to https://github.com/quantumlib/Stim/pull/837 so I could resolve the merge conflicts.

Issue #638, opened 2023-09-22
Remove the constraint that `stim.Tableau.from_state_vector` inputs must be normalized


Strilanc:
It's kind of nit-picky to require that it be a unit vector. Just start by normalizing it.

PR #840, opened 2024-09-27
Remove the constraint that stim.Tableau.from_state_vector inputs must be normalized

drqec:
Just taking a stab at fixing https://github.com/quantumlib/Stim/issues/638. Note, I don't fully understand how Tableau should work but I wrote the tests based on the examples, so I think this should be correct.

- Adds normalization code by scaling each amplitude by the square root of the inverse of the state vector norm
- Adds tests to match the examples from `from_state_vector` including an unnormalized version

### DIFF
--- a/src/stim/stabilizers/tableau.pybind.cc
+++ b/src/stim/stabilizers/tableau.pybind.cc
@@ -2130,16 +2130,8 @@ void stim_pybind::pybind_tableau_methods(pybind11::module &m, pybind11::class_<T
             }
 
             std::vector<std::complex<float>> v;
-            double weight = 0;
             for (const auto &obj : state_vector) {
                 v.push_back(pybind11::cast<std::complex<float>>(obj));
-                weight += std::norm(v.back());
-            }
-            if (weight != 1.0 && weight > 0.0) {
-                std::complex<float> scale = (float)sqrt(1.0 / weight);
-                for (auto &amplitude : v) {
-                    amplitude *= scale;
-                }
             }
 
             return circuit_to_tableau<MAX_BITWORD_WIDTH>(
@@ -2155,7 +2147,7 @@ void stim_pybind::pybind_tableau_methods(pybind11::module &m, pybind11::class_<T
             Args:
                 state_vector: A list of complex amplitudes specifying a superposition. The
                     vector must correspond to a state that is reachable using Clifford
-                    operations.
+                    operations, and can be unnormalized.
                 endian:
                     "little": state vector is in little endian order, where higher index
                         qubits correspond to larger changes in the state index.
--- a/src/stim/stabilizers/tableau.pybind.cc
+++ b/src/stim/stabilizers/tableau.pybind.cc
@@ -2130,8 +2130,16 @@ void stim_pybind::pybind_tableau_methods(pybind11::module &m, pybind11::class_<T
             }
 
             std::vector<std::complex<float>> v;
+            double weight = 0;
             for (const auto &obj : state_vector) {
                 v.push_back(pybind11::cast<std::complex<float>>(obj));
+                weight += std::norm(v.back());
+            }
+            if (weight != 1.0 && weight > 0.0) {
+                std::complex<float> scale = (float)sqrt(1.0 / weight);
+                for (auto &amplitude : v) {
+                    amplitude *= scale;
+                }
             }
 
             return circuit_to_tableau<MAX_BITWORD_WIDTH>(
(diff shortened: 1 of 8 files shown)


src/stim/stabilizers/tableau.pybind.cc
```diff
@@ -2147,7 +2155,7 @@ void stim_pybind::pybind_tableau_methods(pybind11::module &m, pybind11::class_<T
             Args:
                 state_vector: A list of complex amplitudes specifying a superposition. The
                     vector must correspond to a state that is reachable using Clifford
-                    operations, and must be normalized (i.e. it must be a unit vector).
+                    operations.
```
Strilanc, line 2158: Explicitly say it doesn't need to be normalized.

src/stim/stabilizers/tableau.pybind.cc
```diff
@@ -2130,8 +2130,16 @@ void stim_pybind::pybind_tableau_methods(pybind11::module &m, pybind11::class_<T
             }
 
             std::vector<std::complex<float>> v;
+            double weight = 0;
```
Strilanc, line 2133: This should be moved into the actual C++ method being called, instead of only affecting the python API.

src/stim/stabilizers/tableau_pybind_test.py
```diff
@@ -56,6 +54,27 @@ def test_from_named_gate():
...
+        0,
+        0.5**0.5,
+    ], endian='little')
+    assert len(t) == 2
+    assert t.x_output(0) == stim.PauliString("Z_")
+    assert t.x_output(1) == stim.PauliString("_X")
+    assert t.z_output(0) == stim.PauliString("XX")
+    assert t.z_output(1) == stim.PauliString("ZZ")
+
+    unnormalized = stim.Tableau.from_state_vector([
```
Strilanc, line 70: Also test cases that aren't nice numbers, and with intentional small floating point error.

Strilanc (changes requested):
Yeah this is basically the functionality that's desired, but a few more tests and pushing it down into the C++ are needed.

src/stim/stabilizers/tableau.pybind.cc
drqec: I moved it to the underlying C++ method, but after writing the test, I realized normalization already happens within VectorSimulator: https://github.com/quantumlib/Stim/blob/49531a44bb1c412a0186b171e1c0076833707059/src/stim/simulators/vector_simulator.h#L79. So, we don't **need** to explicitly need to pre-normalize the input, or check for zero, but let me know if you'd prefer it that way.

src/stim/stabilizers/tableau_pybind_test.py
drqec: Added fuzz test to C++ layer.

src/stim/util_top/circuit_vs_amplitudes.cc
```diff
@@ -40,15 +40,6 @@ Circuit stim::stabilizer_state_vector_to_circuit(
     }
 
     uint8_t num_qubits = floor_lg2(state_vector.size());
-    double weight = 0;
```
Strilanc, line 43: Hm. Well that's surprisingly simple.

Strilanc (approved):

src/stim/stabilizers/tableau_pybind_test.py
```diff
@@ -57,6 +57,20 @@ def test_from_named_gate():
         stim.Tableau.from_named_gate("X_ERROR")
 
 
+def test_from_state_vector():
```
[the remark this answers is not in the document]
drqec: It seems even if I remove the random scale the following does not hold:

```
t = stim.Tableau.random(n)
t2 = stim.Tableau.from_state_vector(t.to_state_vector(), endian='little')
assert t == t2 # <-- fails
```

`assert stim.Tableau.from_conjugated_generators(\n    xs=[\n        stim.PauliString("+Y"),\n    ],\n    zs=[\n        stim.PauliString("+X"),\n    ],\n) == stim.Tableau.from_conjugated_generators(\n    xs=[\n        stim.PauliString("+Z"),\n    ],\n    zs=[\n        stim.PauliString("+X"),\n    ],\n)`

However if I instead compare `t.to_state_vector() == t2.to_state_vector()` then they are equal, so I could use that instead or try to figure out why t1and t2 aren't equal.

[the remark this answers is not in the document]
drqec: This, however, does work:

```
t = stim.Tableau.random(n)
t2 = stim.Tableau.from_state_vector(t.to_state_vector(endian='little'), endian='little')
np.testing.assert_array_equal(t.to_stabilizers(canonicalize=True), t2.to_stabilizers(canonicalize=True))
```
So it seems `==` is a little tricky because it doesn't canonicalize. Let me know if the above would work.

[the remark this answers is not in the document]
Strilanc: yeah that's fine to canonicalize it

Strilanc:
Thanks for doing this!

PR #844, opened 2024-10-22
Fix exception when sampling with post-selection

cesarBLG:
Calling sinter.collect() for tasks including a postselection_mask raises an exception. This is caused because the decoder predictions are compared to the total number of shots without excluding discarded shots.

### DIFF
--- a/glue/sample/src/sinter/_decoding/_stim_then_decode_sampler.py
+++ b/glue/sample/src/sinter/_decoding/_stim_then_decode_sampler.py
@@ -197,8 +197,8 @@ class _CompiledStimThenDecodeSampler(CompiledSampler):
             raise ValueError("predictions.dtype != np.uint8")
         if len(predictions.shape) != 2:
             raise ValueError("len(predictions.shape) != 2")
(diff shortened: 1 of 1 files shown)


Strilanc:
Could you add a unit test verifying this? I'm surprised it didn't trip existing unit tests to be honest; definitely something that should be tested.

Strilanc (changes requested):
Add a unit test with a postselected task that used to fail but now passes.

Something like taking 1 shot from

```
X_ERROR(1) 0
M 0
DETECTOR rec[-1]
```

cesarBLG:
I added a unit test as suggested.

The reason it didn't trip any test is that circuits in _decoding_test.py are sampled using _sample_decode_helper_using_memory. The post-selection code in that function is not shared with the one in _CompiledStimThenDecodeSampler, so essentially the latter code isn't covered by any test.

perlinm:
What's the status of this PR?  I have the same issue mentioned in the top comment:
```
ValueError: predictions.shape[0] != num_shots
```

PR #862, opened 2024-11-26
add stimcirq tags

emma-louise-rosenfeld:
When a particular cirq operation has a tag, convert it to stim according to the tag's value.

### DIFF
--- a/glue/cirq/stimcirq/_cirq_to_stim.py
+++ b/glue/cirq/stimcirq/_cirq_to_stim.py
@@ -1,7 +1,8 @@
 import functools
 import itertools
 import math
-from typing import Callable, cast, Dict, Iterable, List, Optional, Sequence, Tuple, Type
+from collections.abc import Callable
+from typing import cast, Dict, Iterable, List, Optional, Sequence, Tuple, Type
 
 import cirq
 import stim
@@ -11,7 +12,7 @@ def cirq_circuit_to_stim_circuit(
     circuit: cirq.AbstractCircuit,
     *,
     qubit_to_index_dict: Optional[Dict[cirq.Qid, int]] = None,
-    custom_op_conversion_func: Callable | None = None
+    custom_op_conversion_func: Callable[[cirq.Operation], cirq.Operation] | None = None
 ) -> stim.Circuit:
     """Converts a cirq circuit into an equivalent stim circuit.
 
@@ -36,6 +37,9 @@ def cirq_circuit_to_stim_circuit(
         circuit: The circuit to convert.
         qubit_to_index_dict: Optional. Which integer each qubit should get mapped to. If not specified, defaults to
             indexing qubits in the circuit in sorted order.
+        custom_op_conversion_func: Optional. A function which will transform cirq operators into other cirq operators, to be then 
+            converted to STIM. Useful in e.g. the case of non-Clifford operations in a cirq circuit, which are to be replaced
+            by Clifford operations in STIM.
 
     Returns:
         The converted circuit.
@@ -101,7 +105,7 @@ def cirq_circuit_to_stim_data(
     *,
     q2i: Optional[Dict[cirq.Qid, int]] = None,
     flatten: bool = False,
-    custom_op_conversion_func: Callable | None = None,
+    custom_op_conversion_func: Callable[[cirq.Operation], cirq.Operation] | None = None,
 ) -> Tuple[stim.Circuit, List[Tuple[str, int]]]:
     """Converts a Cirq circuit into a Stim circuit and also metadata about where measurements go."""
     if q2i is None:
@@ -145,21 +149,21 @@ def gate_to_stim_append_func() -> (
     ) -> Callable[[stim.Circuit, List[int]], None]:
         if len(gates) == 1 and not individuals:
             (g,) = gates
-            return lambda c, t: c.append_operation(g, t)
+            return lambda c, t, tag: c.append(stim.CircuitInstruction(g, t, tag=tag))
 
         if not individuals:
 
-            def do(c, t):
+            def do(c, t, tag):
                 for g in gates:
-                    c.append_operation(g, t)
+                    c.append(stim.CircuitInstruction(g, t, tag=tag))
 
         else:
 
-            def do(c, t):
+            def do(c, t, tag):
                 for g in gates:
-                    c.append_operation(g, t)
+                    c.append(stim.CircuitInstruction(g, t, tag=tag))
                 for g, k in individuals:
-                    c.append_operation(g, [t[k]])
+                    c.append(stim.CircuitInstruction(g, [t[k]], tag))
 
         return do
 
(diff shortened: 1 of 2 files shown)


Strilanc (changes requested):
I can't accept this as currently designed.

The basic issue is you are hardcoding your particular usage of tags into the library (replace a gate based on a tag), rather than making a tool that other people would also use. Stim and stimcirq should attach as little semantic meaning to tags as possible; there should be no kind of replacement logic triggered by the presence of a tag. Instead, the conversion should attempt to preserve the tags while making as few changes as possible.

For example, when stimcirq is converting from a stim circuit to a cirq circuit, it could add the tags appearing on the stim instruction to the resulting cirq gates as a string or as a custom class `stimcirq.StimTag`. Then, when converting from cirq to stim, it could look for `stimcirq.CirqTag` tags and turn them into tags in the stim circuit. For handling more general cases (since cirq allows a wide variety of tagging), the user should be able to specify custom tag conversion functions that take a cirq tag and return a stim tag (or vice versa).

emma-louise-rosenfeld:
> I can't accept this as currently designed.
> 
> The basic issue is you are hardcoding your particular usage of tags into the library (replace a gate based on a tag), rather than making a tool that other people would also use. Stim and stimcirq should attach as little semantic meaning to tags as possible; there should be no kind of replacement logic triggered by the presence of a tag. Instead, the conversion should attempt to preserve the tags while making as few changes as possible.
> 
> For example, when stimcirq is converting from a stim circuit to a cirq circuit, it could add the tags appearing on the stim instruction to the resulting cirq gates as a string or as a custom class `stimcirq.StimTag`. Then, when converting from cirq to stim, it could look for `stimcirq.CirqTag` tags and turn them into tags in the stim circuit. For handling more general cases (since cirq allows a wide variety of tagging), the user should be able to specify custom tag conversion functions that take a cirq tag and return a stim tag (or vice versa).

The use case here is not related to the tags in STIM. Instead, here I'm working on the cirq to stim conversion, when the cirq circuit has non-clifford gates. Let's say we are running an injection protocol for example, and we have a parameterized gate in cirq for arbitrary rotations on the injection qubit. Those cirq gates with arbitrary rotations would have a tag that stimcirq understands as a replacement for the non-clifford gates (e.g. a cirq.XPowGate with a tag "H" would become a Hadamard gate in STIM). 

Beyond state injection, this is generally desired functionality since our detector error modeling in pyle is done by converting cirq circuits (e.g. as saved in a pyle datataker) to STIM, and adding noise. In other use cases, for example, we might want to run QEC circuits in pyle on a device with various different qubit frequencies - in which case we need cirq.Detune gates, which in turn, STIM would need to understand as some clifford chosen by the user.  

We could do this conversion in pyle instead if needed, but let me know if you have other ideas on how to handle this.

emma-louise-rosenfeld:
> > I can't accept this as currently designed.
> > The basic issue is you are hardcoding your particular usage of tags into the library (replace a gate based on a tag), rather than making a tool that other people would also use. Stim and stimcirq should attach as little semantic meaning to tags as possible; there should be no kind of replacement logic triggered by the presence of a tag. Instead, the conversion should attempt to preserve the tags while making as few changes as possible.
> > For example, when stimcirq is converting from a stim circuit to a cirq circuit, it could add the tags appearing on the stim instruction to the resulting cirq gates as a string or as a custom class `stimcirq.StimTag`. Then, when converting from cirq to stim, it could look for `stimcirq.CirqTag` tags and turn them into tags in the stim circuit. For handling more general cases (since cirq allows a wide variety of tagging), the user should be able to specify custom tag conversion functions that take a cirq tag and return a stim tag (or vice versa).
> 
> The use case here is not related to the tags in STIM. Instead, here I'm working on the cirq to stim conversion, when the cirq circuit has non-clifford gates. Let's say we are running an injection protocol for example, and we have a parameterized gate in cirq for arbitrary rotations on the injection qubit. Those cirq gates with arbitrary rotations would have a tag that stimcirq understands as a replacement for the non-clifford gates (e.g. a cirq.XPowGate with a tag "H" would become a Hadamard gate in STIM).
> 
> Beyond state injection, this is generally desired functionality since our detector error modeling in pyle is done by converting cirq circuits (e.g. as saved in a pyle datataker) to STIM, and adding noise. In other use cases, for example, we might want to run QEC circuits in pyle on a device with various different qubit frequencies - in which case we need cirq.Detune gates in the cirq circuit, which in turn, STIM would need to understand as some clifford chosen by the user.
> 
> We could do this conversion in pyle instead if needed, but let me know if you have other ideas on how to handle this.

Changed to custom operation functions (see second commit).

glue/cirq/stimcirq/_cirq_to_stim.py
```diff
@@ -5,15 +5,13 @@
...
-import warnings
-
-_STIMCIRQ_TAG_LOOKUP = {"H": cirq.H, "X": cirq.X, "Y": cirq.Y, "Z": cirq.Z}
 
 
 def cirq_circuit_to_stim_circuit(
     circuit: cirq.AbstractCircuit,
     *,
     qubit_to_index_dict: Optional[Dict[cirq.Qid, int]] = None,
+    custom_op_conversion_func: Callable | None = None
```
Strilanc, line 14: Expand the Callable type to clarify the function's intent.

Also add it to the docstring.

glue/cirq/stimcirq/_cirq_to_stim_test.py
```diff
@@ -463,16 +478,15 @@ def test_stimcirq_tags():
...
- REPEAT 3 {
-        H 0
-        X 1
-        TICK
-        M 0 1
-        TICK
-    }
-"""
+        REPEAT 3 {
+                H 0
```
Strilanc, line 485: Do a test where the stim circuit ends up containing a tag.

glue/cirq/stimcirq/_cirq_to_stim_test.py
```diff
@@ -410,7 +410,22 @@ def test_random_gate_channel():
     )
 
 
-def test_stimcirq_tags():
+def test_stimcirq_custom_conversion():
+
+    _tag_lookup = {"H": cirq.H, "X": cirq.X, "Y": cirq.Y, "Z": cirq.Z}
+
+    def _op_conversion(op: cirq.Operation) -> cirq.Operation:
+        """" For converting particular tagged cirq.Operator's to a value described by the tag content.
```
Strilanc, line 418: This docstring isn't relevant to understanding the test

Strilanc (changes requested):

glue/cirq/stimcirq/_cirq_to_stim_test.py
emma-louise-rosenfeld: @Strilanc this required some extra functionality (persisting cirq tags through to stim tags) beyond our use case here but I gave a stab at it. See my most recent commit for an example of what I'm thinking. I haven't run/written tests on it yet.

Strilanc:
Obsoleted by https://github.com/quantumlib/Stim/pull/866

emma-louise-rosenfeld:
> Obsoleted by #866

@Strilanc , https://github.com/quantumlib/Stim/pull/866 does not cover our use case, can you please reopen this?

Strilanc:
Note: resolved offline

PR #864, opened 2024-12-03
Fix observable predictions shape in sinter's MWPF and FussionBlossom decoder

inmzhang:
The packed obs predictions should be of shape `(num_shots, (self.num_obs + 7) // 8)` instead of `(num_shots, self.num_obs)`, which results in error `ValueError: predictions.shape[1] > actual_obs.shape[1] + 1` when simulating codes with multiple logical observables.


Strilanc:
LGTM; would probably be good to add a unit test to the standard battery of tests that exercises this

Strilanc (approved):

PR #888, opened 2025-02-21
Update backtrack_path() to use one less lookup in back_map.

gopinath-vasalamarri:
The initial inputs to (cur_state, prev_state) map to (next, curr) --> which are already computed.   This lets us bypass calling the back_map.at(next) by passing in curr.

### DIFF
--- a/src/stim/search/graphlike/algo.cc
+++ b/src/stim/search/graphlike/algo.cc
@@ -27,16 +27,17 @@
 using namespace stim;
 using namespace stim::impl_search_graphlike;
 
-DetectorErrorModel backtrack_path(const std::map<SearchState, SearchState> &back_map, const SearchState &final_state) {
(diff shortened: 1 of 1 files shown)


Strilanc (changes requested):
Thanks for the submission, but I'm not going to accept this.

First, the backtracking method isn't on the hot path. This has essentially 0 performance impact. Second, the impact would be negative because now you are making copies of the search state (due to using auto instead of const auto & for the type), and this is causing allocations. I also don't like that it's increasing the coupling between the backtracking method and the search method (by requiring a redundant argument to be passed).

A useful version of this change would be to refactor SearchState to use a uint64_t mask, so that no allocations were needed to copy it. This would accelerate the searching process, which is the hot path. Ironically this is how the method used to work, but it was changed because it didn't work for circuits with more than 64 observables. That could be fixed by iterating over the observables 64 at a time and invoking the search for each group of 64. This would be much faster due to not allocating during the hot loop of expanding the search region. If you're interested in speeding up the method, that's what you should do.

That change I would *definitely* accept.

PR #897, opened 2025-03-04
Add workflow to welcome first-time contributors

mhucka:
This adds a workflow that triggers on new issues and PRs. If the author has not contributed to this repository before, the workflow will add a comment with an informative welcome message to the issue or PR.


Strilanc:
I would prefer to wait for lack-of-information to be a problem before slamming people with a wall of text on their first contribution. Stim isn't being overwhelmed by bad contributions, so we don't need to automate this yet IMO.

Strilanc:
Maybe it's just me, but I've always found these wall-of-text welcomes a bit overwhelming, compared to someone showing organically up and commenting on the issue.

mhucka:
No problem.

PR #902, opened 2025-03-17
Add automation to update CITATION.cff upon releases

mhucka:
This GitHub Actions workflow will update the version and date fields in the CITATION.cff file when a new release is done in GitHub.


Strilanc:
Does this do the wrong thing if I do something like release `1.15.0`, then later release `1.14.1` (e.g. due to backporting a fix to a security vulnerability)?

PR #918, opened 2025-03-28
Add bdist_wheel releases of sinter and stimcirq

amaranthjinn:
Use cibuildwheel for sinter and stimcirq so those  two also can be downloaded from Pypi as wheels binary.


amaranthjinn:
test failure pending fix https://github.com/quantumlib/Stim/pull/925/.

.github/workflows/ci.yml
```diff
@@ -168,6 +168,14 @@ jobs:
         with:
           name: "dist-stim-${{ matrix.os_dist.os }}-${{ matrix.os_dist.dist }}-${{ matrix.os_dist.macosarch }}"
           path: dist/*
+      - uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
```
Strilanc, line 171: Why is this line expected to work? cibuildwheel makes stim packages built from C++ code, not sinter or stimcirq packages. And this looks like it's uploading the files from the same directory just under a different name? Won't this just result in releasing stim wheels with the name "sinter"?

If you scroll down to line 185 you will see the tar.gz for sinter being uploaded. I would expect this PR to make some kind of change where it edited that build step and uploaded a wheel there.

I don't really understand the goal here, or why the changes made would achieve the goal.

Strilanc (changes requested):

.github/workflows/ci.yml
amaranthjinn: Thanks for the correction, I misunderstood the scripts.

Strilanc (approved):

Issue #924, opened 2025-04-03
Tests using dev/doctest_proper.py fail with NumPy 2.x due to differences in format of printed numbers


mhucka:
The CI test job `test_sinter` is failing again; this time, it's because PyMatching was updated yesterday to support NumPy 2, which now results in the CI jobs getting NumPy 2 instead of NumPy 1. The `doctest_proper.py` testing approach for code embedded in docstrings was evidently written with NumPy 1.x in mind. The expected results in the docstrings contain numbers printed in plain form like `1.225568`, whereas NumPy 2 change the defqult printed format to (e.g.) `np.float64(1.225568)`. The change breaks the comparisons done by `doctest_proper.py`, resulting in errors like this:

```
*********************************************************************
File ".../glue/sample/src/sinter/__init__.py", line ?, 
in sinter.__test__.sinter.fit_line_slope
Failed example:
    sinter.fit_line_slope(
        xs=[1, 2, 3],
        ys=[10, 12, 14],
        max_extra_squared_error=1,
    )
Expected:
    sinter.Fit(low=1.2928924560546875, best=2.0, high=2.7071075439453125)
Got:
    sinter.Fit(low=np.float64(1.2928924560546875), best=np.float64(2.0),
               high=np.float64(2.7071075439453125))
```

Example of failure in a CI run: https://github.com/quantumlib/Stim/actions/runs/14248758283/job/39941070887?pr=918

PR #925, opened 2025-04-03
Make numpy use legacy printing format for doctests

mhucka:
Works around the problem of #924

The tests of code samples inside docstrings were written with NumPy 1.x in
mind. The expected results contain numbers printed in plain form like
"1.225568", whereas NumPy 2 change the defqult printed format to (e.g.)
np.float64(1.225568). The following is a temporary workaround.


Strilanc (changes requested):
Fix the tests instead of changing the test-running script to make them artificially pass. A doctest is supposed to pass if pasted into an interpreter, and this breaks that concept.

mhucka:
> Fix the tests instead of changing the test-running script to make them artificially pass. A doctest is supposed to pass if pasted into an interpreter, and this breaks that concept.

Due to the fact that this was a difference in Numpy 1 vs 2 behavior, it wasn't immediately obvious to me how to change the examples in the docstrings to work for either case. The correct solution is obviously to figure out where in the code the format difference is originating, but because (a) that would have taken me (someone not familiar with the code) much longer and (b) the check failures were blocking Xiaoxuan's PR, this draft PR was experimenting with a faster temporary solution until one of us could go back and fix the issue properly.

Strilanc:
Obsoleted by https://github.com/quantumlib/Stim/pull/929

I understand you were trying to get a quick workaround, but I really don't like subtly changing the semantics of the ci like this in order to get things to merge. That kind of stuff gets forgotten. I fixed the tests instead.

mhucka:
> Obsoleted by #929
> 
> I understand you were trying to get a quick workaround, but I really don't like subtly changing the semantics of the ci like this in order to get things to merge. That kind of stuff gets forgotten. I fixed the tests instead, which makes this unnecessary.

Yes, and I'm happy to have seen the fix in #929! Thanks for doing that.

PR #929, opened 2025-04-05
Fix sinter doctests

Strilanc:
Make consistent with numpy v2 instead of numpy v1 (triggered by pymatching now supporting numpy v2)

Fixes https://github.com/quantumlib/Stim/issues/924


mhucka:
> Make consistent with numpy v2 instead of numpy v1 (triggered by pymatching now supporting numpy v2)

@Strilanc Does this mean the tests will only succeed in np v2?

PR #943, opened 2025-04-25
connect max_batch_size from CollectionManager to worker RampThrottledSampler

caryan:
When `max_batch_size` is passed to `sinter.collect` it gets passed to the `CollectionOptions` for the `CollectionManager` but then in `CollectionManager.start_workers` it is not passed on to the `RampThrottledSampler` instead `max_batch_shots` is hardcoded to 1024 [here](https://github.com/quantumlib/Stim/blob/9156ab6c97c4f27be1160cd8b6aafb35f7677e57/glue/sample/src/sinter/_collection/_collection_manager.py#L140).  I have seen some small improvements in runtime when I can go for larger batches. 

This PR passes `max_batch_size` to `RampThrottledSampler.max_batch_shots` and adds some test coverage.

### DIFF
--- a/glue/sample/src/sinter/_collection/_collection_manager.py
+++ b/glue/sample/src/sinter/_collection/_collection_manager.py
@@ -129,6 +129,10 @@ class CollectionManager:
     def start_workers(self, *, actually_start_worker_processes: bool = True):
         assert not self.started
 
+        # Use max_batch_size from collection_options if provided, otherwise default to 1024 as large
+        # batch sizes can lead to thrashing
+        max_batch_shots = self.collection_options.max_batch_size or 1024
+
         sampler = RampThrottledSampler(
             sub_sampler=MuxSampler(
                 custom_decoders=self.custom_decoders,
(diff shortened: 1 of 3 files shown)


glue/sample/src/sinter/_collection/_collection_manager.py
```diff
@@ -129,6 +129,10 @@ def __exit__(self, exc_type, exc_val, exc_tb):
     def start_workers(self, *, actually_start_worker_processes: bool = True):
         assert not self.started
 
+        # Use max_batch_size from collection_options if provided, otherwise default to 1024 as large
+        # batch sizes can lead to thrashing
```
caryan, line 133: I pulled this comment from the README comment "setting batch sizes too large causes thrashing" but have not actually profiled myself

Strilanc (approved):
Nice catch

glue/sample/src/sinter/_collection/_collection_manager.py
caryan: For the Github record there is not much to be gained in most situations by moving away from the 4096 default

image

[snippet](https://gist.github.com/caryan/6b941df93b0598e32eef153bd997e8dd)

PR #952, opened 2025-05-08
Add support for converting `OBSERVABLE_INCLUDE` instructions with pauli targets to stimcirq

AlexBourassa:


Strilanc (changes requested):

AlexBourassa:
I made some adjustments to make serialization easier and keep the data type more similar to the other arguments.  I tested it by doing
```python
import pyle.cirqtools.serialization as serialization

sc1 = stim.Circuit("""
    QUBIT_COORDS(5, 5) 0
    R 0
    OBSERVABLE_INCLUDE(0) X0
    TICK
    H 0
    TICK
    M 0
    OBSERVABLE_INCLUDE(0) rec[-1]
    TICK
""")

cc1 = stimcirq.stim_circuit_to_cirq_circuit(sc1)
cc2 = serialization.read_json(json_text=serialization.to_json(cc1))
sc2 = stimcirq.cirq_circuit_to_stim_circuit(cc2)

assert sc1 == sc2
```

And this works fine.  Note that I don't think I can do the serialization round trip as a unit test here because it uses pyle functions.

glue/cirq/stimcirq/_obs_annotation.py
```diff
@@ -51,6 +53,7 @@ def __repr__(self) -> str:
             f'stimcirq.CumulativeObservableAnnotation('
             f'parity_keys={sorted(self.parity_keys)}, '
             f'relative_keys={sorted(self.relative_keys)}, '
+            f'pauli_keys={self.pauli_keys}, '
             f'observable_index={self.observable_index!r})'
         )
```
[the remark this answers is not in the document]
AlexBourassa: Ok, I think it should be fine now.  See the comment below

Strilanc:
> I made some adjustments to make serialization easier and keep the data type more similar to the other arguments. I tested it by doing
> 
> ```python
> import pyle.cirqtools.serialization as serialization
> 
> sc1 = stim.Circuit("""
>     QUBIT_COORDS(5, 5) 0
>     R 0
>     OBSERVABLE_INCLUDE(0) X0
>     TICK
>     H 0
>     TICK
>     M 0
>     OBSERVABLE_INCLUDE(0) rec[-1]
>     TICK
> """)
> 
> cc1 = stimcirq.stim_circuit_to_cirq_circuit(sc1)
> cc2 = serialization.read_json(json_text=serialization.to_json(cc1))
> sc2 = stimcirq.cirq_circuit_to_stim_circuit(cc2)
> 
> assert sc1 == sc2
> ```
> 
> And this works fine. Note that I don't think I can do the serialization round trip as a unit test here because it uses pyle functions.

It shouldn't be necessary to involve pyle functions to make a unit test.

Strilanc (changes requested):
Still needs a test of the *json serialization*

Strilanc:
The json tests are failing. It looks like you didn't update the thing verifyin that the output is what is expected; since you added a field the expectation should be different.

glue/cirq/stimcirq/_obs_annotation_test.py
```diff
@@ -191,9 +191,22 @@ def test_json_serialization():
...
+    )
+    json = cirq.to_json(c)
+    c2 = cirq.read_json(json_text=json, resolvers=[*cirq.DEFAULT_RESOLVERS, stimcirq.JSON_RESOLVER])
+    assert c == c2
+
 
 def test_json_backwards_compat_exact():
-    raw = stimcirq.CumulativeObservableAnnotation(parity_keys=['z'], relative_keys=[-2], observable_index=5)
-    packed = '{\n  "cirq_type": "CumulativeObservableAnnotation",\n  "parity_keys": [\n    "z"\n  ],\n  "observable_index": 5,\n  "relative_keys": [\n    -2\n  ]\n}'
+    raw = stimcirq.CumulativeObservableAnnotation(parity_keys=['z'], relative_keys=[-2], observable_index=5, pauli_keys=["X0", "Y1", "Z2"])
```
Strilanc, line 209: Nope nope nope, do not change the part of the backwards compatibility test that verifies old json will continue to read. Instead add a second value and check that it parses as expected. Like

```
packed_v1 = ...
packed_v2 = ...
raw = ...
assert cirq.read_json(json_text=packed_v1, resolvers=[*cirq.DEFAULT_RESOLVERS, stimcirq.JSON_RESOLVER]) == raw
assert cirq.read_json(json_text=packed_v2, resolvers=[*cirq.DEFAULT_RESOLVERS, stimcirq.JSON_RESOLVER]) == raw
assert cirq.to_json(raw) == packed_v2
```

Strilanc (changes requested):
Getting much closer, just one easily fixed major mistake left.

glue/cirq/stimcirq/_obs_annotation_test.py
AlexBourassa: Ok, I added back the test that reads the old format.  For the new v2 format, I added added an example without the pauli_keys (which should be equal to the old example), and an example with the pauli_keys.  Let me know if I missed anything else!

Strilanc (approved):

PR #961, opened 2025-06-17
Update type checks to include NumPy numeric types

peter-janderks:
`isinstance(np.int64(1), int))` returns False for numpy 2.3.0 while it returns True for numpy 2.2.0 and earlier versions. 
This caused AnonTaskStasts to raise an AssertionError.

### DIFF
--- a/doc/sinter_api.md
+++ b/doc/sinter_api.md
@@ -852,7 +852,7 @@ def to_csv_line(
         >>> print(sinter.CSV_HEADER)
              shots,    errors,  discards, seconds,decoder,strong_id,json_metadata,custom_counts
         >>> print(stat.to_csv_line())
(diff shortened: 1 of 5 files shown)


Strilanc (approved):

Strilanc:
Blergh this probably affects my code in so many places...

Thanks for the fix!

peter-janderks:
I didn't catch other lines that needed to be updated because I didn't actually run the sinter test. 
`pip install . ` didn't work on my apple device running Clang 16. This was fixed by updating to Apple clang version 17.0.0.
The tests passed on my device.

peter-janderks:
I didn't check `python dev/doctest_proper.py --module sinter` 🤦

PR #963, opened 2025-06-24
create a build rule that exposes stim as a pybind_library

NoureldinYosri:
this allows other libraries like https://github.com/quantumlib/tesseract-decoder/ to use stim classes (e.g. DemTarget and DemInstruction) as inputs/outputs directly instead of using `py::object`

this moves the build rule that allowed me to do this from https://github.com/quantumlib/tesseract-decoder/blob/main/external/stim_py.BUILD#L36 to here


Strilanc:
I don't think this is a good idea. The basic problem is that the C++ code is *not* designed to be backwards compatible, and so if the version linked by the installed stim module and linked by whatever you are doing differ from each other than all hell will break loose. It is much **MUCH** safer to only interact with stim's objects via `pybind11::object`, because that part of the API is carefully guaranteed to remain backwards compatible.

For example, pymatching links to the detector error model class instead of using pybind::object. To avoid compatibility issues it actually prints the dem and reparses it, and this has *still* caused various kinds of trouble over time.

PR #964, opened 2025-06-24
Remove unused diagram var

drqec:
Looks like this was originally part of the diagram method, but this diagram `out` var is no longer used here.


drqec:
Pretty sure the test failures aren't related to this PR, but don't see a button to re-run jobs.

PR #982, opened 2025-07-31
Change how pauli_keys are represented in CummulativeObservableAnnotation

AlexBourassa:
I don't love it, but the previous representation is insufficient because if we modify the cirq circuit the "stim qubit id" of these pauli keys gets out of sync.  Therefore, we need to convert the qubit representation to the actual coordinates to keep track of this.

Let me know what you think!  Or if you have suggestions on how to achieve this in another way.

Also regarding the backward compatibility, I added some ugly logic in the __init__ to support `Iterable[str]` to make sure that `test_json_backwards_compat_exact` still passes, but alternatively we could also just avoid that logic and not be backward compatible here.  I don't think we have many (or any?) serialized cirq_circuit with non-deterministic observables yet (certainly none that we care about!).  But anyway let me know what you prefer.


glue/cirq/stimcirq/_obs_annotation.py
```diff
@@ -29,15 +29,31 @@ def __init__(
...
         self.relative_keys = frozenset(relative_keys)
-        self.pauli_keys = frozenset(pauli_keys)
+        _pauli_keys = []
+        for k in pauli_keys:
+            if isinstance(k, str):
+                # For backward compatibility
+                _pauli_keys.append((cirq.LineQubit(int(k[1:])), k[0]))
+            else:
+                _pauli_keys.append(tuple(k))
+        self.pauli_keys = frozenset(_pauli_keys)
```
Strilanc, line 39: What if a user has code `for pauli_key in op.pauli_keys: if k.startswith("..."): ...` ? This would break that.

AlexBourassa: I guess you mean that changing the type of `pauli_keys` from `str` to `tuple[cirq.QiD, str]` if an issue.  The issue is that there are no reference to the "stim qubit id" when we are in cirq, and there is no guarantee that these ids will mean anything useful when we convert the circuit back from stim to cirq.

We could keep the type `str` and do something like `X(1,1)` and then interpret k[1:] as a tuple/Qid.  That would not impact the snippet you showed, but we could certainly come up with another example snippet that would break (for example if the user has code that does `int(k[1:])` it would break...).  I think this kind of string interpretation might be worse, but I'm game if you prefer it? I don't really see a clean way around it... But if you have some preferred format I'm happy to implement it. Fwiw, since this is a pretty recent addition to stimcirq, I don't think it's gotten too much adoption yet, so making a change now is unlikely to impact anyone at this point.

Again, I'm open to whatever you think is the best solution here.  Let me know!

Strilanc: The main thing is that you must make the changes in a way that doesn't break existing code. Existing code is likely to depend on the type of public fields, so you cannot change their type. You can add other fields, though.

glue/cirq/stimcirq/_obs_annotation.py
```diff
@@ -16,7 +16,7 @@ def __init__(
         *,
         parity_keys: Iterable[str] = (),
         relative_keys: Iterable[int] = (),
-        pauli_keys: Iterable[str] = (),
+        pauli_keys: Iterable[tuple[cirq.Qid, str]] | Iterable[str] = (),
```
Strilanc, line 19: You have to use `Union` instead of `|`, because stim supports python versions without | for types.

glue/cirq/stimcirq/_obs_annotation.py
AlexBourassa: Ok, I tried a version of this.  Let me know what you think!

glue/cirq/stimcirq/_obs_annotation.py
AlexBourassa: Done!

glue/cirq/stimcirq/_obs_annotation.py
Strilanc: It looks like this would still break existing code that is expecting the field to be of type Iterable[string].

AlexBourassa: I think the `self.pauli_keys` here should be a `frozenset[str]` (same as before).  Specifically it would contain something like `{'X1' , 'Z2', ...}` just like before.

The input pauli_keys now accepts a two input types:
1. Iterable[str]: Same as before to stay backward compatible
2. Iterable[tuple[Qid, str]]: New (preferred) input since it will associate the pauli key to a cirq qubit (eg `[(cirq.q(0,1), "X1"), (cirq.q(3,2), "Z2"), ...]`)

I don't think this would break existing code.  Although I'm happy to restructure differently (for example we could separate the input into two fields `pauli_keys` and `pauli_keys_qubit`).  Would that be better?

Strilanc: Existing code *consuming* this field may assume that `pauli_keys` is always an iterable of string. That is the breakage I am worried about.

I agree the constructor won't break, as it accepts a superset of what it used to. But for consumption code to remain correct the type must be a subset of what it used to be; not a superset. You likely need to introduce a different field to avoid this problem.

Strilanc (approved):

PR #983, opened 2025-08-02
ReferenceSampleTree Random Access

qec-pconner:
* Adding `operator[]` to `ReferenceSampleTree`.
  * Allows quick access to any reference sample bit within the tree at a given absolute index.
* Updating `src/stim/util_top/reference_sample_tree.test.cc` to check bit positions, as applicable.

### DIFF
--- a/src/stim/util_top/reference_sample_tree.cc
+++ b/src/stim/util_top/reference_sample_tree.cc
@@ -186,6 +186,32 @@ bool ReferenceSampleTree::operator!=(const ReferenceSampleTree &other) const {
     return !(*this == other);
 }
 
+bool ReferenceSampleTree::operator[](size_t index) const {
+    size_t current_absolute_index = 0;
+    bool result;
+    bool value_found = try_get_bit_value(index, current_absolute_index, result);
+    assert(value_found);
+    return result;
+}
+
+bool ReferenceSampleTree::try_get_bit_value(
+    size_t desired_absolute_index, size_t &current_absolute_index, bool &bit_value) const {
+    for (uint64_t k = 0; k < repetitions; k++) {
+        const size_t relative_index = desired_absolute_index - current_absolute_index;
+        if (relative_index < prefix_bits.size()) {
+            bit_value = prefix_bits[relative_index];
+            return true;
+        }
+        current_absolute_index += prefix_bits.size();
+        for (const ReferenceSampleTree &child : suffix_children) {
+            if (child.try_get_bit_value(desired_absolute_index, current_absolute_index, bit_value)) {
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
 std::ostream &stim::operator<<(std::ostream &out, const ReferenceSampleTree &v) {
     out << v.repetitions << "*";
     out << "(";
--- a/src/stim/util_top/reference_sample_tree.h
+++ b/src/stim/util_top/reference_sample_tree.h
@@ -24,6 +24,8 @@ struct ReferenceSampleTree {
     bool operator==(const ReferenceSampleTree &other) const;
     /// Checks if two trees are not exactly the same, including structure (not just uncompressed contents).
     bool operator!=(const ReferenceSampleTree &other) const;
+    /// Returns the bit value for a given absolute index.
+    bool operator[](size_t index) const;
     /// Returns a simple description of the tree's structure, like "5*('101'+6*('11'))".
     std::string str() const;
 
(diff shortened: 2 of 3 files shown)


src/stim/util_top/reference_sample_tree.cc
```diff
@@ -186,6 +186,32 @@ bool ReferenceSampleTree::operator!=(const ReferenceSampleTree &other) const {
...
+    size_t current_absolute_index = 0;
+    bool result;
+    bool value_found = try_get_bit_value(index, current_absolute_index, result);
+    assert(value_found);
+    return result;
+}
+
+bool ReferenceSampleTree::try_get_bit_value(
+    size_t desired_absolute_index, size_t &current_absolute_index, bool &bit_value) const {
+    for (uint64_t k = 0; k < repetitions; k++) {
```
qec-pconner, line 199: I may be able to do this is max 2 iterations, instead of `repetitions` iterations. Will update with new commit soon.

Strilanc: Yeah, if you compute the size of the elements then you can skip ahead to the correct iteration with a division. That will be very important for getting ~O(1) time in practice rather than ~O(n).

Create a test case that encodes a tree with a quadrillion bits (but doesn't decompress it) and tries to index into it. Passing corresponds to it taking less than a milliescond (rather than hours).

qec-pconner: > if you compute the size of the elements then you can skip ahead to the correct iteration with a division.

Yes, but if you actually used `size()` it would always require a full iteration of the tree, even for the first element.
What I changed this to is to do 1 iteration first, which both looks for the element at the desired index (if it happens to be there) and also calculates the size. Then, if the element was not found, we skip forward and (if still in the node) do a second pass to actually find the element. Thus "max 2 iterations" (how that many iterations it actually is in the full recursive tree is a bit more complicated).
I don't think this can be faster without storing the node size as an element (so it's precomputed).

qec-pconner: > Create a test case that encodes a tree with a quadrillion bits (but doesn't decompress it) and tries to index into it.

Did a test with >10^17 bits (encoded into 5 nodes).
Tested a handful of bits from the beginning, the end, and 1 outer repetition before then ending (~1 million bits before the end).
Bazel shows it as taking less than 1ms.

src/stim/util_top/reference_sample_tree.cc
```diff
@@ -149,12 +149,18 @@ ReferenceSampleTree ReferenceSampleTree::simplified() const {
...
-size_t ReferenceSampleTree::size() const {
-    size_t result = prefix_bits.size();
+uint64_t ReferenceSampleTree::size() const {
+    uint64_t result = prefix_bits.size();
     for (const auto &child : suffix_children) {
         result += child.size();
     }
-    return result * repetitions;
+#if defined(__GNUC__) || defined(__clang__)
+    bool overflow = __builtin_umull_overflow(result, repetitions, &result);
```
qec-pconner, line 158: I added this overflow check in to ensure that the size of any node doesn't exceed the `uint64_t` type (which would lead to silent logical bugs, which I ran into when the "large_tree" test accidentally made >10^20 bits at first).

However, I haven't seen much in Stim in terms of checking compilers (for compiler related macros), and I fully admit that `#if` can be gross... so, if you'd prefer, I could just remove this check. I just want to call it out for you to explicitly take a look.

src/stim/util_top/reference_sample_tree.cc
```diff
@@ -149,12 +149,18 @@ ReferenceSampleTree ReferenceSampleTree::simplified() const {
...
+    uint64_t result = prefix_bits.size();
     for (const auto &child : suffix_children) {
         result += child.size();
     }
-    return result * repetitions;
+#if defined(__GNUC__) || defined(__clang__)
+    bool overflow = __builtin_mul_overflow(result, repetitions, &result);
+    assert(!overflow);
+#else
+    result *= repetitions;
```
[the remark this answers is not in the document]
qec-pconner: I came across `_mulx_u64()` (https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mulx_u64) which is not compiler-specific. Replaced with this.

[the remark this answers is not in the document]
qec-pconner: Ya, normally I would suggest having some util class with static methods that hide these `#if` away. Then its just a nice `MathUtil::MultiplyNoOverflow()` call. But, I'm not about to start adding all sorts of general stuff in this PR.

PR #1002, opened 2025-10-23
Fixing sinter post-selection feature

shohamjac:
Starting from sinter 1.14 `sinter.collect` with postselection is raising the following error:
```python
ValueError: predictions.shape[0] != num_shots
```

An example can be seen [here](https://colab.research.google.com/drive/1rAuBMvjVrW6109nELja5vbMB6pt00yjc?usp=sharing).

I have fixed the assertion (`predictions.shape[0] != num_shots - num_discards_1`) and added a post-selection test `test_collect_postselection`.


glue/sample/src/sinter/_collection/_collection_test.py
```diff
@@ -1,4 +1,5 @@
 import collections
+from importlib.metadata import metadata
```
DeDuckProject: is this used anywhere?

Strilanc:
Sorry for letting this sit for so many months. I made a few minor tweaks and will handle merging this in https://github.com/quantumlib/Stim/pull/1057 . Thanks for catching this; again sorry for taking so long.

PR #1014, opened 2025-12-02
Remove extra over-mixing checks

abbrazi:
When processing DEP1 instructions in the ErrorAnalyzer, p > 0.75 is checked in:

a) undo_DEPOLARIZE1 and
b) depolarize1_probability_to_independent_per_channel_probability

Likewise, p > 0.9375 is checked in:

c) undo_DEPOLARIZE2 and
d) depolarize2_probability_to_independent_per_channel_probability

I removed checks (a) and (c). I also added tests for checks (b) and (d).

Let me know if this was intentional.

Signed-off-by: Abbas Bracken Ziad <abbas.ziad25@imperial.ac.uk>

### DIFF
--- a/src/stim/simulators/error_analyzer.cc
+++ b/src/stim/simulators/error_analyzer.cc
@@ -816,9 +816,6 @@ void ErrorAnalyzer::undo_DEPOLARIZE1(const CircuitInstruction &inst) {
     if (!accumulate_errors) {
         return;
     }
-    if (inst.args[0] > 0.75) {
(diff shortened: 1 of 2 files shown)


Strilanc:
Good eye and the change looks correct but unfortunately the redundant check was intentional, in order to give a better error message.

Python users will more easily understand what to do in response to `Can't analyze over-mixing DEPOLARIZE1 errors (probability > 3/4).` than when told `depolarize1_probability_to_independent_per_channel_probability with p>0.75`.

PR #1015, opened 2025-12-04
Fix hyperlink in doc /getting_started.ipynb

BarbieriEdoardo:
This PR fixes the hyperlink in `getting_started.ipynb` that points to the “Additional resources” section. The previous link was missing the leading `#`, so it did not jump to the intended section when clicked.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,17 +29,14 @@ jobs:
       fail-fast: false
       matrix:
         os_dist: [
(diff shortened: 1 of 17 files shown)


doczenwiry:
The problem might also be related to a known issue [PY45804](https://youtrack.jetbrains.com/issue/PY-45804/Internal-hyperlinks-dont-work-in-Notebooks) as the correction you made works in the browser but not in PyCharm.

PR #1034, opened 2026-01-27
Initial commit for stimside

mliqai:
This is the initial commit for Stimside in the glue folder.

Stimside is a stim python wrapper that aims to enable classical state transition, tracking and conditioning that can be used to emulate leakage and loss events on top of either the flip simulator or the tableau simulator in Stim. The flip simulator wrapper is called flipside, and the tableau simulator wrapper is called tableside

It's based on and heavily intertwined with Stim and Sinter.

### DIFF
--- a/glue/stimside/examples/rotated_surface_code.py
+++ b/glue/stimside/examples/rotated_surface_code.py
@@ -42,11 +42,6 @@ class TaskMetadata:
                         return_circuit.append(stim.CircuitInstruction('H', op.targets_copy()[1::2]))
 
                         return_circuit.append(stim.CircuitInstruction(
-                            'I_ERROR', 
-                            op.targets_copy(),
-                            tag = f"LEAKAGE_DEPOLARIZE_1"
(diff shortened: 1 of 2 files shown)


glue/stimside/examples/rotated_surface_code.py
```diff
@@ -0,0 +1,139 @@
...
+        rounds=d,
+        p = p,
+    ).make_task()
+    for d in sweep_ds
+    for p in sweep_ps
+]
+
+## Simulation limits
+max_shots = 1_000_000
+max_errors = 1_000_000
```
m-mcewen: should probably be more like 1000?

mliqai: This is from some earlier benchmarking run. Should we change it to smaller?

m-mcewen: probably just to save people wasting a lot of compute

Issue #1004, opened 2025-10-27
Running shortest_graphlike_distance slow at large circuits/large cycles


gabriellelcr:
Perhaps not a surprise, but stim_circuit.shortest_graphlike_error() is slow at large cycles/large distances (60 seconds for 1_000 rounds at distance 11 for a standard surface code). Is there a way to allow the function to recognize if there's one big loop/some repeating structure, and take less time?

We can get around the timing problem by just running shortest_graphlike_error() on the same circuit with a smaller number of cycles, but it's a convenient feature to be able to run it blindly on any circuit (eg when generating metadata for an experiment).

DeDuckProject:
@gabriellelcr Are you able to reproduce the 60 seconds thing?
I've added some optimization which cut the time by 50%, but the baseline on my machine was never on the order of 60seconds, more like 600ms. So I wonder how your example perform now.

PR #1041, opened 2026-02-25
Optimizing backtrack_path

DeDuckProject:
d25, 25 rounds: 83ms -> 39ms
d11, 1000 rounds: 640ms -> 300ms

This addresses issue #1004, although probably doesn't completely resolve it.
Also, the new benchmark introduces didn't reproduce the 60 seconds in the original issue; it took only ~640ms on my machine.


Strilanc:
Nice and simple. Thanks!

PR #1042, opened 2026-02-26
Add support for comparing DEMs up to instruction ordering

abbrazi:
The current `==` overload of the `DetectorErrorModel` only returns true if the DEMs have the same instruction ordering. It is sometimes useful to be able to compare two DEMs for equality when this is not the case, so I implemented an `is_equal_up_to_instruction_ordering` member function. I opted not to change the `==` operator because (a) checking for equality up to instruction ordering requires sorting so turns an O(n) op into an O(n log n) op; and (b) it's possible some of the backend simulation code depends on the ordering constraint.

### DIFF
--- a/src/stim/dem/detector_error_model.cc
+++ b/src/stim/dem/detector_error_model.cc
@@ -16,9 +16,11 @@
 
 #include "stim/dem/detector_error_model.h"
 
+#include <algorithm> 
 #include <cmath>
 #include <iomanip>
 #include <limits>
+#include <numeric>
 
 #include "stim/util_bot/str_util.h"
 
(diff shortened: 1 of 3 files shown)


src/stim/dem/detector_error_model.cc
```diff
@@ -793,3 +794,86 @@ DetectorErrorModel DetectorErrorModel::without_tags() const {
...
+            index_in_a++;
+            index_in_b++;
+            continue;
+        }
+
+        if (structural_op_a.type == DemInstructionType::DEM_REPEAT_BLOCK) {
+            // check reps
+            if (structural_op_a.repeat_block_rep_count() != structural_op_b.repeat_block_rep_count()) {
+                return false;
+            }
```
abbrazi, line 862: At this point, you might ask why I don't simply check op_a != op_b before handling the dedicated shift/repeat cases. For repeats, that would result in unexpected failures, i.e., where the the DEMs have the same repeat blocks but they are stored in different orders in the backing `blocks` arrays. That seems like an implementation detail that should not result in this member function reporting inequality.

abbrazi:
Have another look at this PR when you get the chance. I updated the ordering-insensitive comparison to handle shifts/repeats efficiently by treating them as structural barriers (must match exactly) and comparing the instruction segments between barriers order-insensitively, so large repeats don’t need to be unrolled.

I still think the most general isomorphic check would be a semantic flatten + canonicalise + sort + compare (where shifts/repeats only matter via their effects), but that’s expensive.

DEM construction seems to be becoming an active area of research, including tools that try to automatically infer detector/logical observable definitions. Utilities like this are useful for validating those approaches against Stim as a golden reference, using an equality notion that’s less strict than == while still being equivalent from the perspective of decoders.

Feel free to reject the PR if you feel these utilities should live elsewhere. I will anyway keep it on my fork.

Strilanc:
I'm worried about the fact that the comparison doesn't correspond to a nice Schelling point. It's a more loose notion of equality, but the things it supports and doesn't support are about implementation rather than concepts the user would really care about. The user might care about "are these encoding approximately the same Tanner graph?" but if they use a method like this as a proxy for that notion then they will be surprised when it fails due to canonicalization steps that are not included. And I'm not sure if the exacting equality check can be implemented efficiently.

Because there seem to be so many choices about what equality a person might want, and because I don't think this hits a particularly common one, I'm going to close the PR. The code looks solid, but I don't want to end up with dozens of notions of equality and including halfway compromises is how I end up with dozens of notions of equality.

abbrazi:
Sure, I got the same impression after implementing the more efficient algorithm. The "are these encoding approximately the same Tanner graph?" is best achieved through flatten+sort+compare, but I appreciate that the slowness of that approach is against the philosophy of this codebase and is likely not worth supporting natively unless there is a significant demand for it.

PR #1050, opened 2026-04-08
Add pan and zoom controls to Crumble

Raycosine:
This adds pan/zoom/scrolling navigation controls to Crumble.

**Keyboard Controls**
- `Arrow Keys (Up/Down/Left/Right)`: Pan the Planar Layout view
- `ctrl+0`: Reset View (reset both the planar view and the timeline scroll)
- `ctrl+-`: Zoom out
- `ctrl++` or `ctrl+=`: Zoom in

**Mouse Controls**
- `shift+MMB` (Middle Mouse Button):
  - Drag on the left (planar layout) to pan the planar layout view
  - Drag on the right to scroll vertically through the wires
- `ctrl+MMB`:
  - Drag vertically on the left (planar layout) to zoom in/out

Fixes #1040

PR #1051, opened 2026-04-08
Added Crumble Zoom+pan & scroll

### LINKED ISSUE #1040: Need for scroll / zoom in / zoom out features in Crumble
The Crumble subproject appears very useful for evaluating logical circuits. 
In case multiple logical qubits are present in the circuit, Stim's QUBIT_COORDS instruction allows for visualizing each logical patch distinctly, and the left side of Crumble editor shows each patch in the 2D space, together with the currently executed logical gate.

Unfortunately, if multiple logical patches need to be displayed, the available space in Crumble editor saturates. Together with @zphy, we were wondering whether it is possible to implement scroll down/up, scroll left/right, zoom in, and zoom out features. That would ease the evaluation of the circuits.

Thanks for your help!

DeDuckProject:
Added Crumble zoom,pan and scroll behaviour:

+ Added zoom/pan for timeslice (left) view (qubit-grid)
+ Added pan (X/Y) for timeline (right) view
+ changed some "magic numbers" to consts (QUBIT_HIGHLIGHT_SIZE, MAX_QUBIT_COORDINATE)
+ Panning activated by middle-button + drag or by scrolling (touchpad)
+ Zoom activated by ctrl/cmd + scroll

zoom_2

Closes #1040


glue/crumble/draw/main_draw.js
```diff
@@ -485,6 +497,7 @@ function draw(ctx, snap) {
     } finally {
         ctx.restore();
     }
+    return maxTimelineScrollY;
```
Strilanc: It's super counter-intuitive that the draw command would return `maxTimelineScrollY`. Provide an alternate route to get at it, like a `getMaxTimelineScrollY` method, and don't return it here.

Alternatively, declare a `DrawSummary` struct with a `maxTimelineScrollY` and return that. And update the docstring of this method to say it's returning that value.

glue/crumble/draw/state_snapshot.js
```diff
@@ -17,8 +17,14 @@ class StateSnapshot {
...
      * @param {!Array<![!number, !number]>} boxHighlightPreview
+     * @param {!number} viewportX
+     * @param {!number} viewportY
+     * @param {!number} viewportZoom
+     * @param {!number} timelineScrollY
+     * @param {!number} curMouseScreenX
+     * @param {!number} curMouseScreenY
      */
-    constructor(circuit, curLayer, focusedSet, timelineSet, curMouseX, curMouseY, mouseDownX, mouseDownY, boxHighlightPreview) {
+    constructor(circuit, curLayer, focusedSet, timelineSet, curMouseX, curMouseY, mouseDownX, mouseDownY, boxHighlightPreview, viewportX=0, viewportY=0, viewportZoom=1, timelineScrollY=0, curMouseScreenX=undefined, curMouseScreenY=undefined) {
```
Strilanc: Why are these new parameters defaulting to values, when the old ones weren't?

glue/crumble/main.js
```diff
@@ -179,14 +190,80 @@ editorState.canvas.addEventListener('mouseup', ev => {
...
+
+function handleTimelineVerticalScroll(ev) {
+    editorState.timelineScrollY += ev.deltaY;
+    restrictQubitGridAndTimeline();
+    editorState.force_redraw();
+    return;
+}
+
+function handleQubitGridZoomPan(ev) {
+    if (ev.ctrlKey || ev.metaKey) {
```
Strilanc, line 272: I think the control scheme should be the following:

1. When you middle-click+drag on either of the views, dragging the mouse should pan. For the timeline view it might be best to keep the X panning locked... but there would be value in the user being able to indicate they want the current layer to be more to the left so they see further into the circuit.

2. When you scroll the mouse wheel while hovering over the timeslice view, it should zoom around the mouse pointer. (Note you can efficiently pan by zooming out with the mouse in one position then zooming in with the mouse in a new position. This, plus the middle-dragging for panning, obsoletes the need for the modifier keys.)

3. When you scroll the mouse wheel while hovering over the timeline view, it should scroll the list of qubits ...or zoom the list. I'm not sure. Scrolling makes sense from a "this is a list" perspective and zooming makes sense from "behave the same as the thing right next to you" perspective. Only way to tell would be to try it and see which felt better.

Strilanc (changes requested):
I really like this functionality, but want to tweak how it's triggered.

glue/crumble/main.js
DeDuckProject: Sure, as for (1), using the middle click+drag does makes sense. Regarding whether the timeline X panning should be locked - the implementation is definitely simpler if we keep it locked, but I agree that allowing for X scroll is something that a user might want. I think I'll give it a go and see how it feels.

(2) that makes sense for the mouse. However, I think it would make using the touchpad (which is what I tested on) unintuitive. I'd argue that using the modifier key for zoom is a better compromise, but I'll give it some more thought tomorrow. What do you think?

(3) if we do decide to go for "scroll is zoom for timeslice" in (2) - then indeed it's not clear whether scrolling for scroll would feel natural here. I will say that with current implementation (scroll is pan) it does feel natural even though it only scrolls the Y axis at the moment.

Strilanc: Hm, I didn't consider using a touchpad. It would also make sense to consider pinch actions etc for users with touch screens on their laptops.

I'm fine with timeslice wheeling meaning to scroll rather than to zoom.

DeDuckProject: Currently pinch works on my touchpad, so I assume it should work similarly on a touch screen - but I'll check.

So to summarize, I think the approach would be:
1. add middle click + drag to pan both panels; As for timeline - ideally we'll add X panning as well (if it's simple enough. We'll probably want to reset X when user moves between layers for focus.
2. Leaving the modifier+scroll for zoom for mouse users; touchpad users can use modifier/pinch for zoom; regular scrolling just pans.
3. since scroll is pan - timeline behaviour will stay the same and feel natural; There will be no zoom in timeline panel.

Raycosine:
> I really like this functionality, but want to tweak how it's triggered.

I would like to suggest keeping the axes anchored to the top and left edges and adjusting the tick intervals while zooming in and out, like what I did https://github.com/quantumlib/Stim/pull/1050. I stole the hotkeys from Blender, though, so the interaction is different from what you had in mind.

Strilanc:
@Raycosine lol I hadn't realized I was getting notifications about two separate PRs.

Keeping the ticks marks anchored does seem like the clearly better behavior.

DeDuckProject:
@Raycosine @Strilanc Yes, freezing the tick marks does make sense. I'll add that after the other stuff.

glue/crumble/draw/main_draw.js
DeDuckProject: Done. added DrawSummary.

glue/crumble/draw/state_snapshot.js
DeDuckProject: Removed default assignments

DeDuckProject:
@Strilanc Ok, I've completed the changes we discussed:

- Panning can now be done mid-click + drag.
- Timeline scroll supports X scroll; resets to selected layer when user clicks next/prev layer (I think that's the expected behaviour, but open to other opinions)
- Both Timeline/Timeslice views freeze the qubit label so they always appear (in the same size).

See updated gif in PR description for demo.
Can merge as far as I'm concerned.

DeDuckProject:
I resolved the conflicts with my previous PR that takes care of the resolution thingie.
Basically switching to using `clientWidth`/`Height` instead of `width`/`height`. Made sure it works as expected.
Screenshot 2026-05-13 at 15 26 34

PR #1052, opened 2026-04-10
Crumble UI touchups

DeDuckProject:
Made some Crumble UI touchups:
- Fixed canvas blurriness which occurs on Retina display (and probably others)
- example circuits open as modal instead of inline (closes by esc/outside-click/clicking the "x")
- html style extracted to <style> tag - still inside the html file to ensure file is self-contained.
- Small improvements to layout (mainly buttons layout)
- main canvas size now dynamically calculated to fit screen (so no more scroll)

After:
Screenshot 2026-04-14 at 0 33 19

Before:
Screenshot 2026-04-10 at 13 38 03


glue/crumble/crumble.css
```diff
@@ -0,0 +1,82 @@
+body {
```
Strilanc: This looks like it's breaking the fact that crumble compiles into a single HTML file (e.g. so that it can be written to stdout which is then piped into a file which is then opened with a browser).

Inline the styles into the HTML file.

glue/crumble/crumble.css
```diff
@@ -0,0 +1,82 @@
...
+    padding-left: 5px;
+    padding-right: 5px;
+}
+table th {
+    border: 1px solid black;
+    padding-left: 5px;
+    padding-right: 5px;
+}
+
+button {
```
Strilanc: Why aren't we just using the default button style? I generally prefer to let each browser choose how widget should look.

Strilanc (changes requested):
I like the retina fixes and the dialog for the example circuits. But the other style changes seem sorta superfluous and appear to be breaking single-file-ness.

(Does hitting escape work for closing the dialog?)

glue/crumble/crumble.css
DeDuckProject: Oh, cool. I didn't realize the inline style was intentional. Thought it only open as URL. will inline the styles, sure.

glue/crumble/crumble.css
DeDuckProject: Just a matter of taste, I guess. I can remove the button style, but still leave the flexbox such that at least they align properly.

DeDuckProject:
> I like the retina fixes and the dialog for the example circuits. But the other style changes seem sorta superfluous and appear to be breaking single-file-ness.
> 
> (Does hitting escape work for closing the dialog?)

I like the escape idea - will implement that + inline styles and revert the button opinionated styling.

DeDuckProject:
@Strilanc, I took care of your comments and can merge as far as I'm concerened.
In particular:
- 'esc' closes dialog
- Brought back CSS into html file, but in a <style> tag, for better separation of concerns
- Removed opinionated styling for buttons, but left the layout (see updated screenshot in the PR description)
- Fixed an issue where opening "import/export" temporarily changed the proportions of the canvas until closing it. Note, we are now calling "applyDevicePixelScaling" from multiple places (when resizing screen, when opening import/export..). Alternatively, we can just listen to the resize of the canvas itself, and handle that in a single place - the issue is that the entire canvas flickers upon resize, so I can add throttling, but I feel all this might be over-engineering and we might as well leave it at that. If you think we should go there - let me know; otherwise let's merge this.

Strilanc (approved):

Strilanc:
LGTM, I'll get it merged. Thanks!

PR #1064, opened 2026-05-13
Add missing include for GCC 16

shumpohl:
Building Stim fails for me on Fedora 44 with GCC 16.

This PR fixes that by adding a missing include.


Strilanc:
Thanks!

( Sure do love different systems disagreeing about which headers are needed... :l I wish there was an easy way to test it. )

Issue #651, opened 2023-11-07
can't delete and insert circuit layer in crumble


peter-janderks:
I'm working on a mac and use chrome. I want to delete and insert circuit layers in crumble. If I press crtl+delete or ctrl-insert nothing happens. All other commands I have tried work.

(As a work around, I just open the stim circuit and insert ticks or delete lines).

Strilanc:
Does it work with `command` instead of `ctrl`?

peter-janderks:
No

Strilanc:
I *think* this was resolved in the meantime? Please reopen if this is still happening; but I've definitely seen people on mac on chrome inserting layers.

nikojpapa:
I am also experiencing this. What are the mac commands to insert and delete for this on mac? `command + delete` and `control + delete` does nothing, and there is no explicit `insert` key.

nikojpapa:
For those tech-savvy, you can simulate the combos in the developer's console (right click on page -> Inspect). Go to the "Console" tab and paste the following code:

```
function dispatchKeyCombo(key1, key2) {
  const element = document.querySelector('#cvn');
  const event1 = new KeyboardEvent('keydown', {
    key: key1,
    ctrlKey: true,
    bubbles: true
  });
  element.dispatchEvent(event1);

  const event2 = new KeyboardEvent('keydown', {
    key: key2,
    ctrlKey: true,
    bubbles: true
  });
  element.dispatchEvent(event2);

  const event3 = new KeyboardEvent('keyup', {
    key: key2,
    ctrlKey: true,
    bubbles: true
  });
  element.dispatchEvent(event3);

  const event4 = new KeyboardEvent('keyup', {
    key: key1,
    ctrlKey: true,
    bubbles: true
  });
  element.dispatchEvent(event4);
}
```

You can now dispatch events like `dispatchKeyCombo("Control", "Insert")` for `ctrl+ins` and `dispatchKeyCombo("Control", "Delete")` for `ctrl_del`.

PR #1065, opened 2026-05-14
Support Command shortcuts in Crumble on macOS

GautamNambiar:
Adds macOS Command-key support for Crumble editing shortcuts.

Changes include:
- Command-click as an alias for Control-click selection toggling
- Command-click support for example links
- Command+Z / Command+Shift+Z / Command+Y for undo and redo
- Command+C / Command+X / Command+v for copy/cut/paste
- Command+Enter as a Mac-friendly insert-layer shortcut
- Command+Backspace / Command+Delete for deleting the current layer
- Toolbox preview suppression while Command is held
- Updated shortcut docs and button labels
- Regenerated embedded Crumble resource

Command shortcuts are handled directly when the key is pressed because, in local macOS testing, routing them through Crumble’s chord preview/finalization flow made shortcuts like Command+Z and Command+Enter unreliable.

Command+V uses the browser paste event instead of directly reading from `navigator.clipboard`, avoiding clipboard-read permission prompts while preserving the existing Control+V path.

Addresses #651.

I worked with ChatGPT to help understand the code and make edits.


GautamNambiar:
The main issue that prompted me to make this fix was that ctrl+click on mac usually leads to the equivalent of a right click in Windows. So I had found difficulty with group-selecting two sets of qubits (as a symmetric difference). The PR fixes it.

glue/crumble/README.md
```diff
@@ -116,15 +116,16 @@ button (now labelled "Hide Import/Export") again.
...
+- `ctrl+delete` or `cmd+delete`: Delete current circuit layer.
 - `ctrl+backspace`: Delete current circuit layer.
 - `ctrl+insert`: Insert empty layer at current circuit layer, pushing current circuit layer ahead in time.
-- `ctrl+z`: Undo
-- `ctrl+y`: Redo
-- `ctrl+shift+z`: Redo
-- `ctrl+c`: Copy selection to clipboard (or entire layer if nothing selected).
-- `ctrl+v`: Past clipboard contents at current selection (or entire layer if nothing selected).
-- `ctrl+x`: Cut selection to clipboard (or entire layer if nothing selected).
+- `cmd+enter`: Insert empty layer at current circuit layer, pushing current circuit layer ahead in time.
```
DeDuckProject:
```suggestion
- `ctrl+insert` or`cmd+enter`: Insert empty layer at current circuit layer, pushing current circuit layer ahead in time.
```

glue/crumble/main.js
```diff
@@ -442,12 +435,136 @@ async function pasteFromClipboard(preview) {
     editorState.commit_or_preview(newCircuit, preview);
 }
 
+/**
+ * @param {!string} text
+ * @param {!boolean} preview
+ */
+function pasteTextFromClipboardEvent(text, preview) {
+    let pastedCircuit = Circuit.fromStimCircuit(text);
```
DeDuckProject, line 406: This is very similar to pasteFromClipboard. This code should be written once.

DeDuckProject:
Description says that Cmd+V behaviour was unchanged, but the code does have it in scope.

glue/crumble/main.js
```diff
@@ -442,12 +435,136 @@ async function pasteFromClipboard(preview) {
...
+    }
+}
+
 const CHORD_HANDLERS = makeChordHandlers();
 /**
  * @param {!KeyboardEvent} ev
  */
-function handleKeyboardEvent(ev) {
+async function handleKeyboardEvent(ev) {
+    if (ev.type === 'keydown' && ev.metaKey) {
```
DeDuckProject, line 476: Wouldn't these better be right next to the other chord handlers (`makeChordHandlers`) with `meta+z` etc.?

glue/crumble/main.js
GautamNambiar: Thanks! This is fixed now. Please also see my reply to the next comment for further clarification.

glue/crumble/main.js
GautamNambiar: I had originally tried what you suggested: `meta+...` chord handlers. But on my browsers (chrome and Safari, macOS), doing so, i.e. using the chord preview/finalization flow made shortcuts like cmd+Z and cmd+Enter unreliable. I don't know why, but handling these shortcuts immediately when the key is pressed made them consistent (this was suggested by chatGPT). It seems that for these shortcuts, bypassing the preview system seems to be harmless. Existing ctrl shortcuts still go through Crumble’s chord handler. 

So I kept the cmd shortcut handling separate from `makeChordHandlers`.

cmd+V has one extra difference: it uses the browser paste event to get pasted text, instead of directly calling navigator.clipboard.readText(). This avoided the browser clipboard-read permission prompt I saw during testing.

ctrl+V still gets text using the existing navigator.clipboard.readText() path.

After the pasted text is obtained, cmd+V and ctrl+V now both call the same pasteTextAtFocus helper, so the actual Crumble editing behavior is shared (so the duplication is fixed).

A short comment has been added above the helper explaining why it exists.

I should note that I know very little about web development. I was asking chatGPT, testing, asking again and so on. So, probably there is a cleaner way to do this...

GautamNambiar:
> Description says that Cmd+V behaviour was unchanged, but the code does have it in scope.

Sorry, my bad. I initially wanted to separate the fix to cmd +v from everything else (because cmd+v was leading to browser permission prompts to look into the clipboard, and so this had to be treated separately). But I messed up with version control.

Anyway, now, cmd+v is in scope and I have updated the description. Please review again if you get a chance. Thanks!

PR #1066, opened 2026-05-14
Support Command+V paste in Crumble on macOS

GautamNambiar:
This PR is stacked on top of #1065 and should be reviewed after that PR.

It adds Command+V paste support in Crumble on macOS using the browser paste event instead of directly reading from `navigator.clipboard`.

This avoids triggering clipboard-read permission prompts while preserving the existing Control+V path.

Note: I used Codex to implement this change, and I tested it locally.


GautamNambiar:
I am closing this because somehow this PR ended up being identical to https://github.com/quantumlib/Stim/pull/1065. I meant to submit the cmd+v part separately, but messed up the version control. https://github.com/quantumlib/Stim/pull/1065 is fine now though.

PR #1077, opened 2026-06-26
Fix "access violation" failures that have appeared in new windows builds by disabling compiler optimization on windows

### LINKED ISSUE #1078: Debug why CI windows builds are suddenly failing with an access violation

Strilanc:
The root cause was a compiler autovectorization bug. Turning off optimizations will hurt performance but at least make the code not *totally broken*.

Reproduction of the bug in a simpler instance:

```
#include <cstdint>
#include <iostream>

uint64_t repro() {
    uint32_t targets[6]{0, 27, 0, 0, 27, 0};
    uint64_t t = 6;
    for (size_t k = 0; k < 6; k++) {
        if (targets[k] == 27) {
            t -= 2;
        }
    }
    return t;
}

int main() {
    std::cerr << "t=" << repro() << "\n";
    return 0;
}
```

Go to godbolt, paste the above code, compile it with the compiler `x64 msvc v19.51 VS18.6` and flags `/O2` and run it and it will print `t=8589934594` instead of `t=2`.

Fixes https://github.com/quantumlib/Stim/issues/1078


0xced:
Excellent [blog post](https://algassert.com/post/2603) on that rabbit hole (I stumbled on it on https://hachyderm.io/@Di4na/116832334461184014)! Although I don't use much python and C++ I feel like you perfectly describe my life as of software engineer. 😆

PR #1092, opened 2026-07-14
fix touched() function in LayerTag

emma-louise-rosenfeld:
Ensures that circuit instructions that are tagged with more than one target group are identified correctly with `def touched()` in `LayerTag` of `stimflow`.

### DIFF
--- a/glue/stimflow/src/stimflow/_layers/_layer_tag.py
+++ b/glue/stimflow/src/stimflow/_layers/_layer_tag.py
@@ -15,18 +15,18 @@ class LayerTag(Layer):
         return LayerTag(circuit=self.circuit)
 
     def touched(self) -> set[int]:  # set of qubit touched by it
-        if isinstance(self.circuit[0], stim.CircuitRepeatBlock):
-            raise NotImplementedError(
-                "No tags allowed on circuit repeat blocks, "
-                "since I need target_groups for defining which qubits are touched"
-            )
-        touched_set = {
-            target.qubit_value
-            for target_group in self.circuit[0].target_groups()
-            for target in target_group
-            if target.is_qubit_target
-        }
-        return {q for q in touched_set if isinstance(q, int)}
+        def _touched(inst: stim.CircuitInstruction | stim.CircuitRepeatBlock) -> set[int]:
+            if isinstance(inst, stim.CircuitRepeatBlock):
+                return {q for inner_inst in inst.body_copy() for q in _touched(inner_inst)}
+            touched_set = {
+                target.qubit_value
+                for target_group in inst.target_groups()
+                for target in target_group
+                if target.is_qubit_target
+            }
+            return {q for q in touched_set if isinstance(q, int)}
+
+        return _touched(self.circuit[0])
 
     def to_z_basis(self) -> list[Layer]:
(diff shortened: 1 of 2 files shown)


glue/stimflow/src/stimflow/_layers/_layer_tag.py
```diff
@@ -15,8 +15,18 @@ def copy(self) -> LayerTag:
         return LayerTag(circuit=self.circuit)
 
     def touched(self) -> set[int]:  # set of qubit touched by it
-        tagged_gate_targets = self.circuit[0].target_groups()[0]
-        return {gate_target.qubit_value for gate_target in tagged_gate_targets}
+        if isinstance(self.circuit[0], stim.CircuitRepeatBlock):
```
Strilanc: It seems like this should iterate through the repeat block rather than failing.

emma-louise-rosenfeld: added a commit for this

glue/stimflow/src/stimflow/_layers/_layer_tag_test.py
```diff
@@ -56,3 +56,32 @@ def test_survives_transpile():
...
+    # We verify touched() handles target groups gracefully.
+    assert isinstance(layer_mpp.touched(), set)
+
+    # Test that when self.circuit[0] is a CircuitRepeatBlock, touched() iterates
+    # through the block and finds all qubit targets.
+    layer_repeat = LayerTag(
+        circuit=stim.Circuit(
+            """
+            REPEAT 5 {
+                CX 0 1
```
Strilanc, line 80: Add a test with a repeat block nested inside another one.

Strilanc (commented):
Looking better, but still has some omissions.

glue/stimflow/src/stimflow/_layers/_layer_tag_test.py
emma-louise-rosenfeld: done

glue/stimflow/src/stimflow/_layers/_layer_tag.py
```diff
@@ -15,8 +15,18 @@ def copy(self) -> LayerTag:
         return LayerTag(circuit=self.circuit)
 
     def touched(self) -> set[int]:  # set of qubit touched by it
-        tagged_gate_targets = self.circuit[0].target_groups()[0]
-        return {gate_target.qubit_value for gate_target in tagged_gate_targets}
+        def _touched(inst: stim.CircuitInstruction | stim.CircuitRepeatBlock) -> set[int]:
```
Strilanc: Linear time instead of quadratic vs nesting level:

```
stack = [inst]
out = set()
while stack:
    cur = stack.pop()
    if isinstance(cur, stim.CircuitRepeatBlock):
        stack.extend(cur.body_copy())
    else:
        for target in cur.targets_copy():
            if target.is_qubit_target:
                out.add(target.qubit_value)
```

emma-louise-rosenfeld: big brain

Strilanc (approved):

PR #300, opened 2022-07-25
Add detector error model sampling with error recording

Strilanc:
Example:

```python
import stim
sampler = stim.DetectorErrorModel("""
    error(0.125) D0
    error(0.25) D1
""").compile_sampler()
det_data, obs_data, err_data = noisy_sampler.sample(shots=100, return_errors=True, bit_packed=True)
```

Python changes:

- Add `stim.DetectorErrorModel.flattened`
- Add `stim.DetectorErrorModel.rounded`
- Add `stim.DetectorErrorModel.compile_sampler`
- Add `stim.CompiledDemSampler` class
- Add `stim.CompiledDemSampler.sample`
- Add `stim.CompiledDemSampler.sample_write`

Doc hanges:

- Fix documented type signatures of `to_file`/`from_file` methods
- Stop listing deprecated methods in the API reference

C++ changes:

- Move more pybind-related methods into the `stim_pybind` namespace
- Add `stim::MeasureRecordReader::read_into_table_with_major_shot_index`
- Add `stim::MeasureRecordReader::read_into_table_with_minor_shot_index`
- Add `stim::read_file_data_into_shot_table`
- Add `stim::RaiiFile(RaiiFile&&)`
- Add `stim::DemSampler`
- Add `stim sample_dem` command line command
- Add `--err_out`, `--err_out_format`, `--replay_err_in`, `--replay_err_in_format` flags for `stim sample_dem`
- Add some cleanup flexibility to `stim::RaiiFile` (manual done, manual donotclose)
- Add `read_contents`/`write_contents` to test utility `RaiiTempNamedFile`
- Change `stim::DetectorErrorModel::rounded` to only round probabilities not coordinates

Fixes https://github.com/quantumlib/Stim/issues/80
Fixes https://github.com/quantumlib/Stim/issues/297

PR #1101, opened 2026-08-15
Add discards_b8_out_path support to sinter file decoders

sdoygb:
Context
#297 identified that Sinter integrations drop the decoder low-confidence status, underestimating the conservative logical failure rate. The compiled-decoder path already signals discards via an extra trailing byte (classify_discards_and_errors in _stim_then_decode_sampler.py), but the file-decoder path (decode_via_files) has no channel for discards.

This PR extends the decode_via_files protocol with an optional discards_b8_out_path parameter (1 byte per shot, nonzero = discard), matching the established trailing-byte semantics. Decoders that don't support the new parameter are detected via inspect.signature and keep the old behavior - zero breakage.

Changes
- _decoding/_decoding_decoder_class.py: optional discards_b8_out_path parameter + docstring; base-class default implementation accepts a trailing discard byte from compiled decoders and forwards it to the discard file.
- _decoding/_decoding.py: _streaming_count_mistakes reads the discard file; _sample_decode_helper_using_disk creates/passes the discard path when supported; _sample_decode_helper_using_memory handles trailing-byte output (previously crashed with a shape mismatch).
- _decoding/_stim_then_decode_sampler.py: DiskDecoder passes the discard path and returns the trailing byte, so the existing classify_discards_and_errors consumes it with no new logic.
- Tests: file decoder with/without discard support (compat), memory-path trailing byte, base-class forwarding, DiskDecoder end-to-end.

Related
- Fixes the file-decoder portion of #297
- Complements danielgaskins' #300 (tesseract-side compiled decoder), which covers the compiled path


sdoygb:
@google-cla — the Individual CLA has now been signed (GitHub account linked). Could you re-run the check? Thanks!

sdoygb:
I have signed the Google Individual CLA and linked my GitHub account. Could the CLA check be re-run? Thanks!

mhucka:
> I have signed the Google Individual CLA and linked my GitHub account. Could the CLA check be re-run? Thanks!

Done. Thank you for your contribution.

Strilanc (changes requested):
I don't think I want to merge this.

There are three problems here: breaking compatibility, complexity ratchet, and performance pitfall.

The breaking compatibility issue is that there are several existing decoders that provide a `sinter.Decoder` implementation, and you have added an argument that they need to implement to "properly" be a `sinter.Decoder`. I realize it's an optional argument and that if it's not specified things might happen to work, but still.

The complexity ratchet issue is that there are many, many things like 'record discards' that people might want to add to a decoder. For example, 'record errors' and 'record corrections' and 'record gaps' and 'record swim distances' and 'record runtime distribution' and 'record details of failed shots' and on and on and on. As we added those things, the notion of a decoder becomes more and more complicated. Eventually it becomes so complicated that the concept becomes unhelpful. I don't want to crank the rachet, and this change cranks the ratchet. (Avoiding the ratchet was the motivation for adding `sinter.Sampler`, which has a simpler less opinionated interface and for which it is far easier for users to hide complicated behavior behind.)

The performance pitfall is that the file writing methods were a mistake from the beginning, because when running many instances on many cores they end up losing parallelism due to contention for disk. It's better to keep things entirely in memory. (This was the motivation for adding the non-file-writing variants (inferred automatically from the file writing one if needed, to avoid breaking backwards compatibility).)

